PR #2, opened 2021-06-18
Phase iii

cwjames1983:
Including changes I have had and tested in local development for the last few weeks.


zdm/tests/test.py
```diff
@@ -67,25 +67,35 @@ def main():
...
 	method=2
 	
+	
+	# sets which kind of source evolution function is being used
+	source_evolution=0 # SFR^n scaling
+	#source_evolution=1 # (1+z)^(2.7n) scaling
+	
+	
+	# sets the nature of scaling with the 'spectral index' alpha
+	alpha_method=0 # spectral index interpretation: includes k-correction. Slower to update
```
cwjames1983, line 85: Probably should change this so the default is alpha_method=1

cwjames1983 (commented):
I have tested these particular functions locally, and aside from having the default alpha_method be 0 in test.py (should probably use the 'rate interpretation'), I think it is all legit

zdm/cosmology.py
```diff
@@ -274,6 +274,39 @@ def dFnu_to_dEnu(z,alpha=0,bandwidth=1.e9):
 	dE = 4*np.pi*(dl(z))**2/(1.+z)**(2.-alpha)
 	return dE
 
+######### possible source evolution functions go here ##########
+
+def choose_source_evolution_function(which=0):
+	"""
```
profxj, line 280: I tend to have a brief `Arguments` list in the doc string detailing
the parameters that may be input to the method.  These defines
their allowed type and a brief note on what they are.

In this example, it would be something like:

```
Arguments:
      which (bool, optional):  defines which function to use.  Default (0) is sfr_evolution
```

profxj: But if you find writing such additional docs to be too
big of a pain, I am fine with that.  And I'll just do it 
from time to time as I touch the code.  :)

zdm/iteration.py
```diff
@@ -848,8 +856,15 @@ def cube_likelihoods(grids,surveys,psetmins,psetmaxes,npoints,run,howmany,outfil
 	
 	# this is the order of fastest to slowest, i.e. we update index
 	# order[0] first, then order [1], etc
-	
-	order=[7,4,5,6,0,1,2,3]
+	# the order can depend on the methods being used
+	# note: takes this from the first grid, it should be identical for all grids
+	if grids[0].alpha_method==0:
```
profxj, line 861: would be great to move from numeric indices for the parameters to string (i.e. easily interpretable).  Just raising this point here, not suggesting we do it now.

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
+
```
profxj, line 1: I might move this to a new folder named `mcmc/`

Typically, `tests` is reserved for "unit tests", i.e. scripts
that test functionality of bits of the code.  Although I realize
that the MCMC codes are in part for that function. :)

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
...
+matplotlib.rcParams['image.interpolation'] = None
+
+defaultsize=14
+ds=4
+font = {'family' : 'normal',
+        'weight' : 'normal',
+        'size'   : defaultsize}
+matplotlib.rc('font', **font)
+
+##### definitions of parameters ########
```
profxj, line 50: After this PR, I am going to refactor the code so that the parameters
are handled with a highly interpretable `pset`.  e.g. a dict!

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
...
+	
+	
+	# sets the nature of scaling with the 'spectral index' alpha
+	alpha_method=0 # spectral index interpretation: includes k-correction. Slower to update
+	#alpha_method=1 # rate interpretation: extra factor of (1+z)^alpha in source evolution
+	
+	############## Initialise surveys ##############
+	
+	# constants of intrinsic width distribution
+	Wlogmean=1.70267
```
profxj, line 98: For this 'fixed' parameters, we should have a separate `dict` that defines them and carries them around.  I can also take it as my task to do so.

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
...
+		pwidths,pprobs=survey.make_widths(pks,Wlogmean,Wlogsigma,Wbins,scale=Wscale)
+		efficiencies=pks.get_efficiency_from_wlist(dmvals,pwidths,pprobs)
+		
+		
+		names=['ASKAP/FE','ASKAP/ICS','Parkes/Mb']
+		
+		surveys=[lat50,ICS,pks]
+		if not os.path.isdir('Pickle'):
+			os.mkdir('Pickle')
+		with open('Pickle/'+sprefix+'surveys.pkl', 'wb') as output:
```
profxj, line 152: Am curious -- is there a way around `pickle`?

These can be tough to handle across platforms..

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
...
+			ICS=surveys[1]
+			pks=surveys[2]
+	print("Initialised surveys ",names)
+	
+	dirnames=['ASKAP_FE','ASKAP_ICS','Parkes_Mb']
+	
+	#### these are hard-coded best-fit parameters ####
+	# initial parameter values. SHOULD BE LOGSIGMA 0.75! (WAS 0.25!?!?!?)
+	# Best-fit parameter values (result from cube iteration)
+	lEmin=30. # log10 in erg
```
profxj, line 169: Would be best if these were input to the method so that
we can vary them in the future.

zdm/zdm.py
```diff
@@ -11,13 +11,32 @@ class grid:
 	It also assumes a linear uniform grid.
 	"""
 	
-	def __init__(self):
+	def __init__(self,source_evolution=0,alpha_method=1,luminosity_function=0):
+		"""
+		Class constructor.
+		Source evolution is the function that determines z-dependence.
```
profxj, line 17: Can you also explain what `alpha_method` is?

profxj (commented):
A few comments for you to chew on @cwjames1983 

I'm fine with your deferring any of the things I mentioned.

zdm/cosmology.py
esanmouli: I can also help with that :)

zdm/cosmology.py
```diff
@@ -274,6 +274,39 @@ def dFnu_to_dEnu(z,alpha=0,bandwidth=1.e9):
...
+	Madau & dickenson 2014
+	Takes only a single param
+	"""
+	return (1.0025738*(1+z)**2.7 / (1 + ((1+z)/2.9)**5.6))**params[0]
+	
+
+def opz_evolution(z,*params):
+	"""
+	Same as SFR, but without numerator, i.e. just (1+z)**2.7
```
esanmouli, line 303: Do you mean without the denominator(instead of the numerator)?

zdm/tests/mc_statistics.py
esanmouli: This is true. I use Windows, and usually, I need to change these parts of the code for them to work on my device.

zdm/tests/mc_statistics.py
```diff
@@ -0,0 +1,386 @@
...
+	
+	savefile='mc_sample.npy'
+	
+	try:
+		sample=np.load(savefile)
+		print("Loading ",sample.shape[0]," samples from file ",savefile)
+	except:
+		Nsamples=10000
+		print("Generating ",Nsamples," samples from survey/grid ",which)
+		sample=g.GenMCSample(Nsamples)
```
esanmouli, line 215: @cwjames1983 I can't seem to find where this new GenMCSample() function is defined. I thought it would be in the zdm.py file which initialises the grid, but it isn't there.

zdm/zdm.py
esanmouli: From what I understood from the code, I think this is basically whether we are using the spectral index interpretation of alpha(which is 0) or the rate interpretation(which is 1)

zdm/zdm.py
```diff
@@ -89,14 +108,23 @@ def calc_dV(self):
...
 		self.sfr_n=n
-		self.sfr=cos.sfr(self.zvals)**n
-	
-	
+		if alpha is not None:
+			self.alpha=alpha
+		#self.sfr=cos.sfr(self.zvals)**n #old hard-coded value
+		self.sfr=self.source_evolution(self.zvals,n)
+		if self.alpha_method==1:
+			self.sfr *= (1.+self.zvals)**(-self.alpha) #reduces rate with alpha
```
esanmouli, line 126: This is for the rate interpretation of alpha right? Why is the self.sfr being multiplied with (1+z)^-𝛼, when in the paper it states that "Under the rate interpretation, the FRB population itself behaves as Φ ∼ Φ(𝑧, 𝜈) = Φ(𝑧)𝜈^𝛼, and the k-correction therefore directly changes the rate, adding an additional factor of (1 + 𝑧)^𝛼 to Eq. 12 (which is Φ(𝑧) ={Φ0/(1 + z)}{SFR(𝑧)/SFR(0)}^n.

esanmouli (approved):
Done reviewing and adding comments/questions.

zdm/cosmology.py
cwjames1983: I'll try to gradually add these in the code, and with future commits. Given that this will be an ongoing task for some time however, I'll ask that this gets accepted and I'll gradually fix things.

zdm/iteration.py
cwjames1983: I would have thought perhaps a dict? e.g. params["Emin"]? But in future, I want to divide this up "parameters" into different groups. One defining the FRB luminosity function, since some LFs will need different params than just Emin, Emax, gamma (e.g. Ebreak1,gamma1, Ebreak2,gamma2 etc); we currently have two params defining the host galaxy contribution, but this could be generalised; etc.

zdm/tests/mc_statistics.py
cwjames1983: Well, this is currently only a mc, rather than a mcmc... but please feel free!

zdm/tests/mc_statistics.py
cwjames1983: OK, yes, dict it is!

zdm/tests/mc_statistics.py
cwjames1983: In theory the width could be included in the general parameters. From looking at the output, both width, and the host DM properties, are not strongly correlated with source evolution or luminosity function, and I could imagine in an mcmc fitting the whole lot together. So putting this in a dict, then having 'cube' take arguments to loop over parameters as "Emin 40 42 11" or some such where it looks up "Emin" in the dict would make sense. So I would not make a separate dict for this one.

zdm/tests/mc_statistics.py
cwjames1983: I don't know of a good way to save class objects with various internal arrays other than 'pickle', except manually saving e.g. each thing as various numpy arrays and loading individually.
Btw, does pickle just not work on Windows at all, or does what you read back in not equal what you put out? If you pickle on what platform and re-load on another this is expected to cause problems, but I'm surprised it's an issue within a platform. In any case, this pickle behaviour is fundamentally a speedup, so we might try making a universal flag titled "DoPickle" or some such that if turned to False just ignores the pickle load/save and generates everything fresh.

zdm/tests/mc_statistics.py
cwjames1983: As in lEmin is loaded at the input stage? Yes, agreed. But we would also want the pickling to be passed along. Probably the program should be: EITHER: load a pickled grid and extract params from there, OR: generate a new grid with given parameters.

zdm/zdm.py
cwjames1983: Esan is correct. alpha_method=0 is assuming FRBs have a spectrum as F(nu)~nu**-alpha (NOTE: the paper has nu**alpha, but the code still has the original convention!), with Emax essentially fixed at some nu0; in this case, one uses a k-correction for FRBs at redshift z. alpha_method=1 is the 'rate interpretation', where you assumed the spectral index measured by J-P is due to more narrow-band FRBs at low frequency, so while Emax stays constant, as you go to higher redshift, you end up with fewer FRBs. The rate interpretation is quicker to evaluate because when you cycle over alpha you do not have to recalculate your luminosity function.

zdm/cosmology.py
cwjames1983: Oops, yes, good catch!

zdm/tests/mc_statistics.py
cwjames1983: Wait it's not? It is on the local version I pushed! OK, I will see what the issue is.

zdm/tests/mc_statistics.py
esanmouli: Right, so path names and locations in Windows use the backslash \ instead of the normal slash /. This creates a problem wherein I usually need to go to the specific part in the code where the path is specified and make some other changes as well. I don't think this is specific to pickle, just a problem when using the same code across platforms. The universal flag idea does sound good.

zdm/zdm.py
esanmouli: Got it!

zdm/tests/mc_statistics.py
cwjames1983: I am deciding to leave this in tests for now, it is a relatively small-scale program, and makes sense as a 'test' - it is absolutely NOT a full MCMC!

zdm/iteration.py
cwjames1983: I have decided this is a future to-do, and may get messy, so Im leaving this for now, and will work on it once Esan has a stable version going.

zdm/tests/mc_statistics.py
cwjames1983: Ah OK! So it's essentially any part of the code that has a hard-coded '/' in it. I wonder how this is normally dealt with? A global variable called "appropriate_slash"

zdm/tests/mc_statistics.py
cwjames1983: Since this is an extra feature, unrelated to this pull request, I'll leave this for the moment.

zdm/cosmology.py
cwjames1983: Fixed...

zdm/zdm.py
cwjames1983: Hi Esan - because the original definition in the paper used F(nu)~nu^-alpha. Then it was suggested to change the definition in the paper, but I did not dare tinker with the code at that point. It's a to-do.

esanmouli: Okay, got it.

zdm/tests/mc_statistics.py
profxj: the 'easy' way is to use `os.path.join`

e.g.  `os.path.join('Pickle', filename)`

PR #3, opened 2021-07-06
Branch h0

esanmouli:
Adding H0 as a parameter and a test file.


zdm/H0_func_and_tests.py
```diff
@@ -0,0 +1,254 @@
+# -*- coding: utf-8 -*-
```
profxj, line 1: I suggest the tests go in the `tests/` folder.

zdm/iteration.py
```diff
@@ -16,10 +17,10 @@
 #### this verson currently works with 5 params.
 # soon to be upgraded with two values of local DM contribution
 
-def get_likelihood(pset,grid,survey,norm=True,psnr=False):
+def get_likelihood(pset,grid,survey,norm=True,psnr=True):
```
profxj, line 20: can you add to the doc string here, i.e. define the arguments including `psnr`
thx

zdm/iteration.py
```diff
@@ -28,35 +29,49 @@ def get_likelihood(pset,grid,survey,norm=True,psnr=False):
 		ng=1
 		ns=1
 	if ng==1:
-		update_grid(grid,pset)
+		update_grid(grid,pset,survey)
 		if survey.nD==1:
-			loglik=calc_likelihoods_1D(grid,survey,pset,norm=norm,psnr=psnr)
+			#loglik=calc_likelihoods_1D(grid,survey,pset,norm=norm,psnr=psnr)
```
profxj, line 34: remove

zdm/iteration.py
```diff
@@ -28,35 +29,49 @@ def get_likelihood(pset,grid,survey,norm=True,psnr=False):
...
+			llsum,lllist,expected=calc_likelihoods_1D(grid,survey,pset,norm=norm,psnr=True,dolist=1)
 		elif survey.nD==2:
-			loglik=calc_likelihoods_2D(grid,survey,pset,norm=norm,psnr=psnr)
-		
+			#loglik=calc_likelihoods_2D(grid,survey,pset,norm=norm,psnr=psnr)
```
profxj, line 37: remove -- same for all the other old lines of code

zdm/iteration.py
```diff
@@ -194,19 +215,41 @@ def update_grid(grid,pset):
...
 	oldgamma=grid.gamma
 	oldsfrn=grid.sfr_n
 	oldsmean=grid.smear_mean
 	oldssigma=grid.smear_sigma
+	oldH0value=grid.H0
 	new_sfr_smear=False #do not need to calculate new product of sfr and smearing
 	
+	##check if H0 value changed
+    
+    if oldH0value != pset[8]:
```
profxj, line 240: maybe use `H0value` instead of `pset[8]` for easier comprehension

zdm/misc_functions.py
```diff
@@ -2017,7 +2023,7 @@ def plot_grid_2(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,name='tem
 	dz=zvals[1]-zvals[0]
 	if norm==1:
 		zDMgrid /= ddm
-		if Aconst:
+		if Aconts:
```
profxj, line 2026: this seems like a typo, i.e. `Aconst` seems correct.

zdm/pcosmic.py
```diff
@@ -82,15 +83,18 @@ def make_C0_grid(zeds,F):
 		C0s[i]=iterate_C0(z,F)
 	return C0s
 
-def get_mean_DM(zeds):
+def get_mean_DM(zeds,H0=cos.DEF_H0):
 	""" Gets mean average z to which can be applied deltas """
 	
 	zmax=zeds[-1]
 	nz=zeds.size
 	sys.path.insert(1, '/Users/cjames/CRAFT/FRB_library/FRB-master/')
```
profxj, line 91: this hard-coding is dodgy at best. but maybe it is ignored..

profxj (changes requested):
Great stuff.  Mainly cosmetic requests from me.

zdm/misc_functions.py
esanmouli: The function `plot_grid_2` defines a parameter `Aconts=False`, so I had to change this to` Aconts.` Perhaps the typo is in the function then, and it should be `Aconst=False` ?

zdm/iteration.py
esanmouli: Yep, will do.

zdm/iteration.py
esanmouli: Okay, will do.

zdm/pcosmic.py
esanmouli: This is so that 67.74 isn't hard-coded everywhere and the default value set in the cosmology.py file is taken. Wasn't sure how else to do this.

zdm/iteration.py
esanmouli: Yep, should've done that

zdm/iteration.py
cwjames1983: Agreed, this should be the default

zdm/iteration.py
cwjames1983: OK, so H0 is now added at the end of the parameter set? Can you double check there are no lines with pset[-1] anywhere in the code? I might have used this somewhere to refer to the constant.

zdm/H0_func_and_tests.py
```diff
@@ -0,0 +1,254 @@
...
+lat50.process_survey_file(sdir+'parkes_mb_class_I_and_II.dat')
+DMhalo=50
+lat50.init_DMEG(DMhalo)
+lat50.init_beam(nbins=Nbeams[0],method=2,plot=False,thresh=thresh) # tells the survey to use the beam file
+pwidths,pprobs=survey.make_widths(lat50,Wlogmean,Wlogsigma,Wbins,scale=Wscale)
+efficiencies=lat50.get_efficiency_from_wlist(dmvals,pwidths,pprobs)
+    
+weights=lat50.wplist
+#misc_functions.plot_efficiencies(lat50)
+
```
cwjames1983, line 62: The functionality from lines 62 to 132 seems to be a combination of what is already within "initialise_grids" in "misc_functions.py", plus plots. I would suggest moving this to one call to "initialise grids" and for you to make another routine called "make_lots_of_plots". This shortens the bulk of the code into a few function calls and makes it easier to work out what is going on. If there is something in "initialise_grids" that needs a modification, then we should update that!

zdm/cosmology.py
```diff
@@ -29,7 +29,7 @@
 DEF_Omega_b=0.044
 DEF_Omega_b_h2=0.0224 #Planck says 0.0224, WMAP 0.02264
 # hubble constant in current epoch
-DEF_H0 = 67.4 #km s^-1 Mpc^-1 #planck
+DEF_H0 = 67.74 #km s^-1 Mpc^-1 #planck15 used in frb.igm
```
cwjames1983, line 32: Can we instead read this directly from the FRB library? We should to remove hard-coded values. So e.g. load frb definitions and define DEF_H0 as the value from there

zdm/iteration.py
esanmouli: Okay.

zdm/iteration.py
esanmouli: Ah okay, I will do that! Will make the necessary changes. Thanks!

zdm/H0_func_and_tests.py
esanmouli: Yes, most of this is actually inbuilt into the initialise_grids function - I just didn't use that in the beginning and diden't change it later. I should actually go back and use that function now. In my local update to this code, I have indeed added another "plots" routine to make the code look more manageable. I apologize for the mess here.

zdm/cosmology.py
esanmouli: Okay, will do!

zdm/iteration.py
```diff
@@ -28,35 +29,49 @@ def get_likelihood(pset,grid,survey,norm=True,psnr=False):
...
+			#loglik=calc_likelihoods_2D(grid,survey,pset,norm=norm,psnr=psnr)
+			llsum,lllist,expected=calc_likelihoods_2D(grid,survey,pset,norm=norm,psnr=True,dolist=1)
+		return llsum,lllist,expected
+		#negative loglikelihood is NOT returned, positive is.	
 	else:
 		loglik=0
 		for i,g in enumerate(grid):
 			s=survey[i]
-			update_grid(g,pset)
+			update_grid(g,pset,s)
```
cwjames1983, line 45: Out of interest, what information does update_grid need from this survey?

zdm/iteration.py
```diff
@@ -194,19 +215,41 @@ def update_grid(grid,pset):
...
 	oldsmean=grid.smear_mean
 	oldssigma=grid.smear_sigma
+	oldH0value=grid.H0
 	new_sfr_smear=False #do not need to calculate new product of sfr and smearing
 	
+	##check if H0 value changed
+    
+    if oldH0value != pset[8]:
+        cos.set_cosmology(H0=H0value)
+        zDMgrid, zvals,dmvals,H0=misc_functions.get_zdm_grid(H0=H0value,new=True,plot=False,method='analytic')
```
cwjames1983: One issue with this implementation is that, for multiple grids, it will generate and load a new zDM grid. In theory however, this could be done and set constant for a range of grids. That is, this component of the zdm grid will be identical for ASKAP ICS and FE, and Parkes. Can you find a way to do this only once? Or alternately, outside this function, you can use the zdm grid from one 'grid' object and pass it to another one? In any case, see if you can find a speed-up.

zdm/iteration.py
```diff
@@ -194,19 +215,41 @@ def update_grid(grid,pset):
...
+        grid.pass_grid(zDMgrid,zvals,dmvals,H0)
+        grid.smear_mean=oldsmean
+        grid.smear_sigma=oldssigma
+        grid.smear_dm(oldmask,oldsmean,oldssigma)
+        grid.calc_thresholds(survey.meta['THRESH'],survey.efficiencies,alpha=oldalpha, weights=survey.wplist)
```
cwjames1983: The weights from survey.wlist shouldn't change I don't think. Thresholds will, yes, but can you recycle the wplist from the grid being passed? If the grid does not store survey.efficiencies or survey.wplist then this is fine I guess. Just wondering. Clearly you have to call "calc_thresholds" again. E.g. re-use "weights=grid.eff_weights" should work. Same with efficiencies=grid.eff_table. I don't think efficiencies or eff_weights changes with H0?

zdm/misc_functions.py
```diff
@@ -402,6 +402,8 @@ def make_dm_redshift(grid,savename,DMmax=1000,zmax=1,loc='upper left',Macquart=N
 	
 	plt.legend(loc=loc)
 	plt.savefig(savename)
+	plt.title("H0 " + str(H0))
+    plt.show()
```
cwjames1983: "plt.show" is infinitely annoying when not part of some option to the function - having plots come up on screen automatically drives me nut. Can you at least put the plt.show inside some kind of
if show:
     plt.show()
 and have "show=False" by default?

zdm/misc_functions.py
```diff
@@ -1644,10 +1646,11 @@ def plot_1d(pvec,lset,xlabel,savename):
 	plt.plot(pvec,lset)
 	plt.tight_layout()
 	plt.savefig(savename)
+	plt.show()
```
cwjames1983, line 1649: ditto... :-)

zdm/misc_functions.py
```diff
@@ -1761,11 +1764,11 @@ def get_zdm_grid(new=True,plot=False,method='analytic',
...
 		plt.close()
-		
-		
-	return zDMgrid, zvals,dmvals
 
-def plot_zdm_basic_paper(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,name='temp.pdf',ylabel=None,label='$\\log_{10}p(DM_{\\rm EG},z)$',project=False,conts=False,FRBZ=None,FRBDM=None):
+	#return zDMgrid, zvals,dmvals    
+    return zDMgrid, zvals,dmvals, H0
+
+def plot_zdm_basic_paper(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,name='temp.pdf',ylabel=None,label='$\\log_{10}p(DM_{\\rm EG},z)$',project=False,conts=False,FRBZ=None,FRBDM=None,title='Plot',H0=cos.DEF_H0):
```
cwjames1983, line 1771: Can you take away the H0 part in the function call, since I think you only use it to define the title, and since "title" gets passed anyway, somebody could just call the function with title=old_title+"_"+str(H0)

zdm/misc_functions.py
```diff
@@ -1901,10 +1904,12 @@ def plot_zdm_basic_paper(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,
 	plt.tight_layout()
 	
 	plt.savefig(name)
+	plt.title(title+str(H0))
+    plt.show()
 	plt.close()		
 
 
-def plot_grid_2(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,name='temp.pdf',label='$\\log_{10}p(DM_{\\rm EG},z)$',project=False,conts=False,FRBZ=None,FRBDM=None,Aconts=False,Macquart=None):
+def plot_grid_2(zDMgrid,zvals,dmvals,zmax=1,DMmax=1000,norm=0,log=True,name='temp.pdf',label='$\\log_{10}p(DM_{\\rm EG},z)$',project=False,conts=False,FRBZ=None,FRBDM=None,Aconts=False,Macquart=None,title="Plot",H0=cos.DEF_H0):
```
cwjames1983, line 1912: ditto re the title! Also a quick heads-up: in papers, you almost never add a title to a plot, you just describe in with a caption. Absolutely fine to have a title option, but something to keep in mind for the future paper.

zdm/misc_functions.py
cwjames1983: Esan is corect -  "Aconts" is referring to "contours" for "conts" - i.e. it was likely originally a typo. Perhaps that function was deprecated and I just never used it, it should have failed every time...

zdm/misc_functions.py
```diff
@@ -2246,6 +2253,7 @@ def plot_efficiencies(survey,savename='Plots/efficiencies.pdf'):
 	plt.legend(loc='upper right',fontsize=min(14,200./survey.NFRB),ncol=ncol)
 	plt.tight_layout()
 	plt.savefig(savename)
+	plt.show()
```
cwjames1983, line 2256: Usual comment with all the plt.show().!

zdm/pcosmic.py
cwjames1983: This seems fine to me - but note my comment on how to get DEF_H0 in the first place!

zdm/iteration.py
esanmouli: Along with the grid and pset, because we are recalculating thresholds the `update_grid` function requires the survey file as well because `survey.meta['THRESH'], survey.efficiencies and survey.wplist` are required for `calc_thresholds`

zdm/iteration.py
esanmouli: The `update_grid` function only works for a single grid right? Ex- in `get_likelihood` we are using a enumerate function to loop over every grid. And anyway creating the `zDMgrid, zvals, dmvals, H0` takes about **0.2 to 0.3 seconds** on my computer - shouldn't be too much of a problem! I could instead use the zdm grid from one grid object and pass to the other one. Let me know if I need to!

zdm/iteration.py
esanmouli: No, none of them change! I am indeed reusing the `wplist, efficiencies`, etc. However they are stored in the `survey` object, **not** the `grid `object. This is actually why `update_grid()` is taking another parameter which is the `survey` itself.

zdm/misc_functions.py
esanmouli: Sure, I will make this change right away. Sorry for that! My IDE shows the plots in a different pane, so it wasn't a problem for me.

zdm/misc_functions.py
esanmouli: I could do that, but then the `H0` information gets lost - if this function called the `grid` then I could easily extract the `H0 `value from it. However since it calls` zDMgrid, zvals` and `dmvals` there is no way to figure out what the value of` H0` is!

zdm/misc_functions.py
esanmouli: Okay, will keep that in mind! I will change these then...it's just that the titles help me when I'm actually obtaining the plots

profxj (approved):
approving and merging!

PR #5, opened 2021-09-14
Updated code to allow for lower frequency ASKAP ICS observations

cwjames1983:
Hey all,

I'll try to find a telecon to quickly review what I had to do here. Essentially:
- added new FRB data in Surveys
- added new beamshape data for lower-frequency ASKAP obs
- removed an incorrect 100 MHz offset in the central frequency of ASKAP ICS observations (will have a super tiny effect on submitted paper)
- added functionality to account for frequency offsets between instruments (was not actually in the code would you believe it!) for both alpha methods

Cheers,
Clancy


zdm/beams.py
```diff
@@ -27,14 +27,21 @@ def gauss_beam(thresh=1e-3,nbins=10,freq=1.4e9,D=64,sigma=None):
 		# uses 1.2 lambda on D
 		# check with Parkes: 1.38 GHz at 64m is 14 arcmin
 		HPBW=1.22*(constants.c/(freq*1e6))/D
-		sigma=(HPBW/2.)*(2*np.log(2))**0.5
+		sigma=(HPBW/2.)*(2*np.log(2))**-0.5
```
cwjames1983, line 30: this was a bug! This part of the code never got used, but still, best to change it...

zdm/beams.py
```diff
@@ -48,15 +55,8 @@ def load_beam(suffix,basedir=beams_path):
 	if logb.size == omega_b.size+1:
 		# adjust for the bin centre
 		db=logb[1]-logb[0]
-		logb=logb[:-1]+db
+		logb=logb[:-1]+db/2.
```
cwjames1983, line 58: minor bug fix: logb should be in the centre of the bin, this made it at the top of the bin. This bug will have resulted in artificially increased sensitivity by approximately 1%.

zdm/beams.py
```diff
@@ -27,14 +27,21 @@ def gauss_beam(thresh=1e-3,nbins=10,freq=1.4e9,D=64,sigma=None):
...
-		sigma=(HPBW/2.)*(2*np.log(2))**0.5
+		sigma=(HPBW/2.)*(2*np.log(2))**-0.5
 	# this gives FWHP=0.23 deg = 13.8 arcmin i.e. agrees with Parkes
 	#print("basic deg2 over range 1e-3 is ",2*np.pi*sigma**2*(180/np.pi)**2*6.9)
 	omega_b=np.full([nbins],2*np.pi*dlnb*sigma**2) #omega_b per dlnb - makes sense
 	return b,omega_b
 
 def load_beam(suffix,basedir=beams_path):
+	"""
+	Retrieves beam data.
```
profxj, line 38: add a small comment on what `suffix` and `basedir` are supposed to be.

zdm/beams.py
profxj: good catch

zdm/cosmology.py
```diff
@@ -247,13 +247,15 @@ def E_to_F(E,z,alpha=0, bandwidth=1e9):
 
 
 # inverse of above
-def F_to_E(F,z,alpha=0, bandwidth=1e9):
+def F_to_E(F,z,alpha=0, bandwidth=1e9, Fobs=1.3e9, Fref=1.3e9):
 	""" Converts a fluence to an energy
 	Formula from Macquart & Ekers 2018
 	Bandwidth in Hz
```
profxj, line 253: add a few comments on the arguments, i.e. define what they are.

zdm/data/Surveys/CRAFT_ICS_892.dat
```diff
@@ -0,0 +1,15 @@
+BW    336 #MHz
```
profxj, line 1: Just a reminder that this Repo is public.
So be mindful of what shows up in files like these.

profxj (approved):
Put in a few requests for doc string additions.

And a warning on public vs. private FRBs..

zdm/data/Surveys/CRAFT_ICS_892.dat
profxj: yes, fraid so.

profxj (approved):
all good.  merging

PR #1, opened 2021-04-02
Phase I

profxj:
Gets the code up and running

Have successfully run `test.py` which is now in `tests`

PR #7, opened 2021-11-02
Refactor grids and a fair bit more

profxj:
Tested against cubes in the `test_main_cube` branch.

C speed-up tested but not implemented.


cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
+""" JXP VERSION """
+
+'''
+Questions for Clancy:
+
+1. lC vs. C
```
cwjames1983, line 6: lC. For cubing, it has absolutely no preference at all, in the sense that the numerical minimiser searching for the optimum C can begin with a pretty good guess. However, when implementing some generic minimum fining algorithm, I expect it would be easier to operate on the log10. Also keep in mind that I am keeping with log10 for all logs here.

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
...
+2. Order of update_grid()
```
cwjames1983, line 7: This will require a detailed review. Provided changing H0 is kept as an outermost loop, naively changing everything for H0 (i.e. effectively creating a new grid, albeit in such a way that python does not leak memory - oh, for c++! - won't lose much time). In the future we should aim to improve speed here. Also for evaluating the energy function - there are opportunities to use more memory and save time in the long run.

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
...
+3. Survey is unique to a grid, right?
```
cwjames1983, line 8: Yes

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
...
+4. Order of grid parameters
+5. Why 90s per iteration?
+6. Is it ok to get a NAN??   And why is it ignored??
```
cwjames1983, line 11: This is a bug. It is not meant to be ignored, it's because a NAN is generated later anyway. It is OK to get a NAN, since this means "you have 0 chance of detecting this FRB with these parameters". if we wanted to be really smooth about this, we would search for the closest non-zero point in zdm space to that FRB, and assign a large but non-infinite penalty proportional to that distance, to ensure that any minimiser sees a slope to minimise against.

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
...
+7. https://docs.scipy.org/doc/scipy/reference/tutorial/integrate.html#faster-integration-using-low-level-callback-functions
```
cwjames1983, line 12: In a magical future where I don't need to sleep and there are 50 hr per day I will read this.

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
+""" JXP VERSION """
+
+'''
+Questions for Clancy:
+
+1. lC vs. C
+2. Order of update_grid()
+3. Survey is unique to a grid, right?
+4. Order of grid parameters
```
cwjames1983, line 9: I believe this is resolved.

cubes/new_cube.py
```diff
@@ -0,0 +1,254 @@
...
+5. Why 90s per iteration?
```
cwjames1983, line 10: When testing this, the main conclusion was that we are performing calculations over large grids of O~1 million points. I tried speeding this up with... I forget which pre-compiled c library code I tried - and it didn't help, the conclusion being that numpy is highly optimised already for large grid operations.

zdm/grid.py
```diff
@@ -0,0 +1,672 @@
...
+        to dictate how far each tree should proceed
+        in calculation.
+        Direct variable inputs are always listed first
+        We see that sfr evolution and dm smearing
+        lie just before the pdv step
+        Hence, we deal with these first, and
+        calc rates as a final step regardless
+        of what else has changed.
+        
+        calc_rates:
```
cwjames1983, line 499: We should update this help to reflect the new structure with H0. I am OK making this a to-do for now - perhaps leave a note in the code to this effect however?

zdm/grid.py
```diff
@@ -0,0 +1,672 @@
...
+            new_sfr_smear=True  # True for either alpha_method
+        if self.chk_upd_param('alpha', vparams, update=True):
+            set_evol = True
+            calc_pdv = True
+            new_pdv_smear=True
+            if self.state.FRBdemo.alpha_method == 0:
+                calc_thresh = True
+            elif self.state.FRBdemo.alpha_method == 1:
+                new_sfr_smear=True
+        if set_evol:
```
cwjames1983, line 567: I think we should re-order this routine into two halves (it is almost there anyway):
#1: perform all the checks for changes parameters, and set changes to True/False accordingly
#2: run through all the IF (changes) THEN (calculate) steps
We are close, but there are a couple of tings out-of-order below

zdm/grid.py
```diff
@@ -0,0 +1,672 @@
...
+
+        # TODO -- Thresholds and pdv come *before* in initialize_grids
+        #  SHOULD WE DO THE SAME HERE??
+
+        ##### examines the 'pdv tree' affecting sensitivity #####
+        # begin with alpha
+        # alpha does not change thresholds under rate scaling, only spec index
+        if calc_thresh:
+            self.calc_thresholds(
+                self.F0,self.eff_table, bandwidth=self.bandwidth,
```
cwjames1983, line 578: e.g. here we are calculating new thresholds before deciding if we need to calc_pdv below at line 584. I'd suggest swapping these for "niceness" reasons.

cwjames1983: PS - I fully realise X that I am asking you to nicen up the mess I originally created. Umm... thanks! :-)

zdm/parameters.py
```diff
@@ -112,7 +110,7 @@ class IGMParams(myDataClass):
         default=0.32,
         metadata={'help': 'F parameter in DM PDF for the Cosmic web'})
 
-# Host parameters
+# Host parameters -- host
 @dataclass
 class HostParams(myDataClass):
```
cwjames1983, line 115: I would have expected host parameters to include the log width too. What am I missing?

zdm/tests/test_basic.py
```diff
@@ -17,80 +18,43 @@
...
-    #DEF_Omega_b_h2=0.0224 #Planck says 0.0224, WMAP 0.02264
-    # hubble constant in current epoch
-    #DEF_H0 = igm.Planck15.H0.value #km s^-1 Mpc^-1 #planck15 used in frb.igm
+    # Variable parameters
+    vparams = {}
+    vparams['cosmo'] = {}
+    vparams['cosmo']['H0'] = 67.74
+    vparams['cosmo']['Omega_lambda'] = 0.685
+    vparams['cosmo']['Omega_m'] = 0.315
+    vparams['cosmo']['Omega_b'] = 0.044
```
cwjames1983, line 29: Question here. Do we have a method to fix either H0 or Omega_b, and automatically set the other one according to Planck? I know Esan is implementing this, not sure if it is contained here or not.

zdm/tests/test_basic.py
```diff
@@ -17,80 +18,43 @@
...
 
+    # Update state
+    state.update(vparams)
 
     ############## Initialise cosmology ##############
-    cos.set_cosmology(params)
+    cos.set_cosmology(state)
```
cwjames1983, line 35: can you make the variable "deep_state"? ;-)

zdm/tests/testlib.c
```diff
@@ -0,0 +1,6 @@
+#include <math.h>
+
+/* testlib.c */
+double f(int n, double *xx){
+    return xx[3]*exp(-0.5* pow((xx[0]-xx[1])/xx[2],2));
+}
```
cwjames1983, line 6: What is that stop sign symbol at the end - it that an ascii character? Just checking my editors will be able to make sense of it. Or is this something that github viewer is adding?
Anyway I highly approve of c++, I never used python until this job.

zdm/iteration.py
```diff
@@ -313,7 +319,7 @@ def update_grid(grid,pset,survey):
         grid.rates=grid.pdv*grid.sfr_smear #does pdv mult only, 'by hand'
     
 
-def calc_likelihoods_1D(grid,survey,pset,doplot=False,norm=True,psnr=False,Pn=True,dolist=0):
+def calc_likelihoods_1D(grid,survey,lC,doplot=False,norm=True,psnr=False,Pn=True,dolist=0):
```
cwjames1983, line 322: right, because the grid knows its own pset now. Understood. Might see if there is some way to separate out the lC normalisation step in future, but for now we'll leave this floating here. Might not want to ither. probably not worth the mental effort right now.

zdm/iteration.py
```diff
@@ -835,7 +851,8 @@ def missing_cube_likelihoods(grids,surveys,todo,outfile,norm=True,psnr=True,star
     f.close()
 
 
-def cube_likelihoods(grids,surveys,psetmins,psetmaxes,npoints,
+def cube_likelihoods(grids,surveys, #psetmins,psetmaxes,npoints,
```
cwjames1983, line 854: I'm going to state right now that even in my implementation I needed to run this to check it was behaving properly - viewing the code didn't help me much, so while I'm ticking this one off, it will really need its own test.

zdm/iteration.py
```diff
@@ -869,55 +886,62 @@ def cube_likelihoods(grids,surveys,psetmins,psetmaxes,npoints,
...
     
-    NPARAMS=8
-    nsets=len(grids)
-    lls=np.zeros([nsets])
+    NPARAMS = len(vparam_dict)
+    PARAMS = list(vparam_dict.keys())
+    nsets = len(grids)
+    lls = np.zeros([nsets])
     
     f=open(outfile,'a+')
```
cwjames1983, line 904: We probably want it to output a binary file eventually. Or asap. Then we provide a reader which loads the param class so we know the length and the order. Again, a non-essential to-do.

zdm/iteration.py
```diff
@@ -965,39 +996,55 @@ def cube_likelihoods(grids,surveys,psetmins,psetmaxes,npoints,
...
+            print(f"vparams: {vparams}")
+
             ### minimise if appropriate ### 
             if minimise:
-                C_ll,C_p=my_minimise(pset,grids,surveys,disable=disable,psnr=True,PenParams=None,Verbose=False)
-                toset=np.where(active==1)
-                for s in toset:
-                    pset[s]=C_p[s]
+                #C_ll,C_p=my_minimise(pset,grids,surveys,
+                C_ll = my_minimise(vparams,grids,surveys,
```
cwjames1983, line 1009: one might consider that the minimise functionality is deprecated. But let's keep it here for now.

cwjames1983 (commented):
Hey X - can you elaborate on what dm_file.npz is? I couldn't find where it is needed it text, and understood we are aiming to avoid loading up computer-generated data to github.
I think this is the only potential change here - will approve once you have commented on this.

zdm/grid.py
profxj: made a few edits

zdm/grid.py
profxj: done

zdm/grid.py
profxj: moved

zdm/tests/test_basic.py
profxj: yes for `H0`

zdm/tests/test_basic.py
profxj: :)

zdm/tests/testlib.c
profxj: must be a GitHub thing

zdm/iteration.py
profxj: agreed.  I might implement it.

profxj:
The `dm_file.npz` was just for testing.  It is not needed to run the code.

cwjames1983 (approved):
All good! Will consider your question asked via email (or was it slack?) shortly

zdm/grid.py
```diff
@@ -0,0 +1,665 @@
...
+            smear_mask = True
+            smear_dm = True
+            new_sfr_smear=True
+
+
+        # SFR?
+        if self.chk_upd_param('sfr_n', vparams, update=True):
+            set_evol = True
+            new_sfr_smear=True  # True for either alpha_method
+        if self.chk_upd_param('alpha', vparams, update=True):
```
cwjames1983, line 579: if alpha_method==0, we need
calc_pdv=True
new_pdv_smear=True
calc_thresh=True [since thresholds change]
if alpha_method==1,  we only need to set
set_evol=True
new_sfr_smear=True [since effectively this is changing source evolution]

zdm/grid.py
```diff
@@ -0,0 +1,665 @@
...
+        reset_cos, get_zdm, calc_dV = False, False, False
+        smear_mask, smear_dm, calc_pdv, set_evol = False, False, False, False
+        new_sfr_smear, new_pdv_smear, calc_thresh = False, False, False
+
+        # Cosmology -- Only H0 so far
+        if self.chk_upd_param('H0', vparams, update=True):
+            reset_cos = True
+            get_zdm = True
+            calc_dV = True
+            smear_mask = True
```
cwjames1983, line 559: Changing H0: this does not require the line
smear_mask=True
since these are simply the host galaxy properties that depend only on z, and the mask should remain unchanged.

zdm/grid.py
```diff
@@ -0,0 +1,665 @@
...
+            # we do not NEED the following, but it records this info 
+            # for faster computation later
+            #self.sfr_smear_grid=np.multiply(self.smear_grid.T,self.sfr).T
+            #self.pdv_sfr=np.multiply(self.pdv.T,self.sfr)
+        
+    def calc_thresholds(self, F0, eff_table, bandwidth=1e9, nuObs=1.3e9, nuRef=1.3e9, weights=None):
+        """ Sets the effective survey threshold on the zdm grid
+        F0: base survey threshold
+        eff_table: table of efficiencies corresponding to DM-values
+        nu0: survey frequency (affects sensitivity via alpha - only for alpha method)
```
cwjames1983, line 269: nu0 is meant to say nuObs, oops...

zdm/grid.py
profxj: got it and this is now matching what was
in `update_grid`

zdm/grid.py
profxj: cleaned up

PR #9, opened 2021-11-29
Turn on H0

profxj:
As titled.  And the refactoring continues.
Main changes:

- Enable cube runs with refactored code
- Cleaner handling of parameter order
- Bug fixes here and there
-  Introduce gamma function for energetics
-  Added analysis codes for cubes
-  Added tests using CRACO FRBs
- `pcosmic` speed-up using C code


cubes/new_cube.py
```diff
@@ -60,8 +61,17 @@ def main(Cube):
     # Location for maximisation output
     outdir='Cube/'
 
+    #psetmins,psetmaxes,nvals=misc_functions.process_pfile(Cube[2])
+    input_dict=io.process_jfile(Cube[2])
```
cwjames1983, line 65: So the cube being passed lives in a json file? I have no idea what the general format of that is. The file above (H0_lite_cube.json) looks human readable - but is this adding further complexity?

cubes/new_cube.py
```diff
@@ -189,8 +199,6 @@ def main(Cube):
         if not os.path.exists(outdir):
             os.mkdir(outdir)
         
-        #psetmins,psetmaxes,nvals=misc_functions.process_pfile(Cube[2])
-        vparam_dict=misc_functions.process_jfile(Cube[2])
         run=Cube[0]
```
cwjames1983, line 202: Can you add some comments on whether this is cubing over multiple dimensions or just a single vector by default? i.e. is this syntax general or for only 1D?

cubes/new_cube.py
```diff
@@ -240,7 +248,8 @@ def main(Cube):
 main(Cube)
 
 '''
-python new_cube.py -n 1 -m 3 -p all_params.json -o tmp.npy --clobber
+python new_cube.py -n 1 -m 3 -p all_params.json -o tmp.out --clobber
+python new_cube.py -n 1 -m 3 -p H0_params.json -o H0.out --clobber
 starti is  0
 cosmology: CosmoParams(H0=67.66, Omega_k=0.0, Omega_lambda=0.6888463055445441, Omega_m=0.30966, Omega_b=0.04897, Omega_b_h2=0.0224178568132, fixed_H0=67.66, fix_Omega_b_h2=True)
 FIX THIS!!!!!
```
cwjames1983, line 255: What exactly needs to be fixed? they need to be updated? provided those .json files are easy to read/write by a human this seems fine. Do they need a special method to read them?

cwjames1983: Also: what do n and m do? Is the idea that this is how you *want* it to run?

docs/nb/Exploring_Emax.ipynb
```diff
@@ -0,0 +1,641 @@
+{
```
cwjames1983, line 1: So I have not generally worked with ipython notebooks. From seeing the python code I can't make much of this. Is the ideas that these are demonstrative notebooks that you have developed during your working? I can see these as being worthwhile. But we must ensure that these are kept up-to-date so we don't end up with too much stuff on "main" that does not work.

zdm/analyze_cube.py
```diff
@@ -0,0 +1,477 @@
...
+import pandas
+
+from matplotlib import pyplot as plt
+
+from zdm import io
+from zdm import iteration
+
+from IPython import embed
+
+def slurp_cube(input_file:str, prefix:str, outfile:str,
```
cwjames1983, line 19: Nice routine to have! Big improvement on what I wrote

zdm/analyze_cube.py
```diff
@@ -0,0 +1,477 @@
...
+        # Check
+        if debug:
+            embed(header='69 of analyze')
+    
+    # Write
+    np.savez(outfile, ll=ll_cube, lC=lC_cube)
+    print(f"Wrote: {outfile}")
+
+
+def get_bayesian_data(lls:np.ndarray,
```
cwjames1983, line 78: OK, I clearly need to extend this to 2D in the future.

zdm/analyze_cube.py
```diff
@@ -0,0 +1,477 @@
...
+            pickle.dump(uvals, output, pickle.HIGHEST_PROTOCOL)
+            pickle.dump(vectors, output, pickle.HIGHEST_PROTOCOL)
+            pickle.dump(wvectors, output, pickle.HIGHEST_PROTOCOL)
+        
+    # result is just the total probability, normalised to unit, when summed over the parameter space
+    # technically needs to be divided by the x-increment in bins.
+    return uvals,vectors,wvectors
+
+
+def do_single_plots(uvals,vectors,wvectors,names,tag=None, fig_exten='.png',
```
cwjames1983, line 159: FYI, a lot of this routine is affected by fluff I put in to make the plots more publication ready, but maybe leave for now? Another thing to do might be to extract all the useful info, save it, and then do the plotting in another routine? Anyway, only possible to-dos.

zdm/iteration.py
```diff
@@ -1020,16 +863,15 @@ def cube_likelihoods(grids,surveys, #psetmins,psetmaxes,npoints,
...
                 vparams['lC']=C
 
             
             # for the minimised parameters, set the values
             for j,n in enumerate(current):
-                #string +=' {:8.2f}'.format(pset[j])
                 string +=' {:8.2f}'.format(vparams[PARAMS[j]])
             
             
+            # TODO -- Should we do this?
```
cwjames1983, line 874: Hmm, need to think about this, I suspect not given my previous comment. Might need to have a parameter returned from a previous routine.

zdm/iteration.py
```diff
@@ -852,21 +720,24 @@ def missing_cube_likelihoods(grids,surveys,todo,outfile,norm=True,psnr=True,star
     f.close()
 
 
-def cube_likelihoods(grids,surveys, #psetmins,psetmaxes,npoints,
+def cube_likelihoods(grids:list,surveys:list,
```
cwjames1983, line 723: Hey @profx as a demonstrator to this, it'd be good to have a quick run where the cube routine does not update the grid (i.e. it runs very quickly!) but the cubelikelihoods *does* iterate and spit out the parameters of each iteration. Just to ensure it is behaving properly. I assume you've done a check of this behaviour in suitably many dimensions?

zdm/survey.py
```diff
@@ -103,18 +104,24 @@ def process_survey_file(self,filename):
...
-        #except:
-        self.NFRB=keys.count('FRB')
+
+        # 
+        if NFRB is None:
+            self.NFRB=keys.count('FRB')
+        else:
+            self.NFRB = NFRB
         if self.NFRB==0:
             raise ValueError('No FRBs found in file '+filename) #change this?
```
cwjames1983, line 114: For now - leave it there. In the future, we want to re-write this such that you can have a survey with no results, just survey parameters. Although clearly such a survey could not be fit to, one could clearly create a zdm grid for it (provided a mean Galactic DM component was used)

zdm/survey.py
```diff
@@ -70,7 +71,7 @@ def get_efficiency_from_wlist(self,DMlist,wlist,plist,model="Quadrature"):
         self.mean_efficiencies=mean_efficiencies #be careful here!!! This may not be what we want!
         return efficiencies
     
-    def process_survey_file(self,filename):
+    def process_survey_file(self,filename, NFRB=None):
```
cwjames1983, line 74: Why do we want to tell it what NFRB is? Is the idea that we only wish to *use* NFRB out of a possibly greater number in the file? Can you you write down what the use of this parameter is?

zdm/survey.py
```diff
@@ -475,7 +494,7 @@ def load_survey(survey_name:str, state:parameters.State, dmvals:np.ndarray,
     # Do it
     srvy=Survey()
     srvy.name = survey_name
-    srvy.process_survey_file(os.path.join(sdir, dfile))
```
cwjames1983, line 478: You will soon find a conflict with one of my pull requests, where I take the last "else" to assume the Survey file is "[survey_name].dat" (although I do not test to see if such a file exists, due to laziness). This is so one does not have to change the code when using a new survey file. I don't know how conflicting merges will be handled here for the two concurrent pull requests...

zdm/tests/craco.py
```diff
@@ -0,0 +1,102 @@
+""" CRACO FRBs: This may move out of tests someday """
```
cwjames1983, line 1: My main comment is that this file *should* be redundant with my craco mc stats file, since the first step there is to set up CRACO and generate ZDM grids. I propose that these two files get merged into a single functionality at the next update. Let me know if you or I should do this.

zdm/tests/mc_analysis.py
```diff
@@ -0,0 +1,845 @@
+#!/usr/bin/env python3
```
cwjames1983, line 1: OK this is HUGE! Is this Esan's work? Can you add a couple more comments at the top of the file describing what this is? (not that I do this enough... just do what I say, not what I do! ;-) )

zdm/tests/test_with_craco.py
```diff
@@ -0,0 +1,154 @@
+""" Run tests with CRACO FRBs """
```
cwjames1983, line 1: Aha! So this is the routine you have used to generate your 1D plots? Great! Maybe we can upload this, I'll upgrade to this version, and I'll first try to reproduce your strange H0 plot, then I'll try to debug those jumps

zdm/zdm.py
```diff
@@ -112,31 +113,74 @@ def vector_diff_power_law(Eth,*params):
...
+    """
+    params=np.array(params)
+    Emin=params[0]
+    Emax=params[1]
+    gamma=params[2]
+
+    # Calculate
+    norm = Emax*float(mpmath.gammainc(gamma, a=Emin/Emax))
+    Eth_Emax = Eth/Emax
+    # If this is too slow, we can adopt scipy + recurrance
```
cwjames1983, line 138: So scipy + recurrence means ? Would this mean first precalculating this integral on a grid of a, storing it, and using lookup tables? Or else one could treat the integral as two parts; one, for a <= a* << 1, where it is essentially a pure power law and hence trivially fast; and for a > a* where you care about it being a gamma function.

cwjames1983 (changes requested):
Hey @profxj not sure so much of "changes", but a couple of comments would be good to respond to, in particular the NFRB input to "survey" and the read survey file method defaulting to survey_name.dat if the survey name is not explicitly listed.

Also, adding the 5000 CRACO FRB file seems like putting too much derived data on GitHub. I think we want to avoid that in general. Or where do draw the line? Thoughts?

Btw thanks for all the development you have been doing here, this is great!

cubes/new_cube.py
profxj: Yes, the info defining the cube resides in these JSON files.

The good and bad of JSON files is that they follow a rigid format:
Good -- you don't have to think much at all about reading and writing them
Bad -- you have to conform to them (I guess you can call that complexity)

cubes/new_cube.py
profxj: This follows exactly what we had before,
which I'll admit I don't fully follow.  So, 
I'm not sure what to comment here. ;)

cubes/new_cube.py
profxj: `n` and `m` are the same as before.

I've removed the `FIX THIS` comment.  It was old.

And, there is a new method to read the JSON files.  In `zdm/io.py`

docs/nb/Exploring_Emax.ipynb
profxj: These are only for simple development and testing.

They are *not* part of the code base.  And therefore
need not be kept up to date.

zdm/analyze_cube.py
profxj: thx

zdm/analyze_cube.py
profxj: :+1:

zdm/analyze_cube.py
profxj: yeah, let's leave for now.

zdm/iteration.py
profxj: I think it is ok and am leaving it for now.

zdm/iteration.py
profxj: Let's discuss.  Not sure I follow.

zdm/survey.py
profxj: Roger

zdm/survey.py
profxj: yes, you have that right.  Have added docs.

zdm/survey.py
profxj: Got it.  I think things will behave ok.

But, I've set it so you have to provide `Nbeam` in this case.

zdm/tests/craco.py
profxj: I've taken the initiative to bring together all the CRACO files in a `craco/` folder.

zdm/tests/mc_analysis.py
profxj: It is @esanmouli 's work.

Not sure all it entails! :)

zdm/tests/test_with_craco.py
profxj: yes, that's right.  And I have moved it into the `craco/` folder and renamed it `testing.py`

zdm/zdm.py
profxj: I somewhat forget.  It refers to a comment on the intranets for speeding
up this type of integral.  Probably isn't worth it.

cwjames1983 (approved):

PR #10, opened 2021-12-12
Prepping towards proper H0 runs

profxj:
As titled.  Specifically:

- Added a speed-up for multiple surveys
- Ran tests on CRACO with a Gamma function for energies


zdm/craco/MC_Surveys/CRACO_alpha1_Planck18_Gamma.dat
```diff
@@ -0,0 +1,5015 @@
+BW    288 #MHz
+FRES  1 #MHz
+NFRB 1 #Number of FRBs
+DIAM 12
+NBEAMS 36
+BEAM lat50_log #prefix of beam file
+FBAR 1320
+TRES  1.7 #ms
+SNRTHRESH 9.5
+NFRB 100
```
cwjames1983, line 10: Make it 5000? Same for NORM_FRB

zdm/grid.py
```diff
@@ -379,6 +389,11 @@ def GenMCSample(self,N,Poisson=False):
             which w bin it generates.
         
         """
+        # Boost?
```
cwjames1983, line 392: Can you adda comment about precisely what "boost" means? Is this the cpp code?

cwjames1983: IGNORE the above - sadly, the webpage is not letting me delete the comment for some inane reason...

zdm/grid.py
```diff
@@ -628,26 +656,41 @@ def update(self, vparams:dict, ALL=False):
...
-            zDMgrid, zvals,dmvals=misc_functions.get_zdm_grid(
-                self.state, new=True,plot=False,method='analytic')
-            # TODO -- Check zvals and dmvals haven't changed!
-            self.pass_grid(zDMgrid,zvals,dmvals)
+            if prev_grid is None:
+                zDMgrid, zvals,dmvals=misc_functions.get_zdm_grid(
+                    self.state, new=True,plot=False,method='analytic')
+                self.pass_grid(zDMgrid,zvals,dmvals)
+            else:
+                # Pass a copy (just to be safe)
```
cwjames1983, line 669: Agreed that copying is safer, I came unstuck by this at one point during development and it was a pain.

zdm/grid.py
```diff
@@ -628,26 +656,41 @@ def update(self, vparams:dict, ALL=False):
...
                              self.state.host.lsigma), self.zvals)
+            else:
+                self.smear = prev_grid.smear.copy()
         if smear_dm or ALL:
-            self.smear_dm(self.smear)
+            if prev_grid is None:
+                self.smear_dm(self.smear)
+            else:
+                self.smear = prev_grid.smear.copy()
+                self.smear_grid = prev_grid.smear_grid.copy()
```
cwjames1983, line 693: OK, I really like this ability to just copy the necessary info, it should make it much easier to use multiple grids. We might want to chase up in future whether or not to use "copy" here, since a bunch of these should remain untouched once created, and the difference could be significant for very many FRBs - but we're not at that point yet.

zdm/grid.py
```diff
@@ -406,7 +421,16 @@ def GenMCFRB(self,pwb=None):
             only the relative probabilities of any given width.
             Hence, this routine only returns the integer of the width bin
             not the width itelf.
-        
+
+        Args:
+            pwb (optional): ??
```
cwjames1983, line 426: PWB stands for "probability(width,beam)". It is the total probability of detecting an FRB with a given width at a given beam sensitivity. It takes a lot to calculate, so having it stick in memory and only having it get recalculated once no matter how many FRBs are needed is very useful. This routine first decides which b, w to sample from. We could/should extend this memory to keeping a "rate" grid in memory for every single b,w - but this might be too memory intensive (O~5x5x1000x1000).

zdm/iteration.py
```diff
@@ -1623,12 +1626,32 @@ def minus_poisson_ps(log10C,data):
...
+                        Verbose=True, use_prev_grid:bool=True):
+    """
     Only minimises for the constant, but returns the full likelihood
     It treats the rest as constants
     the grids must be initialised at the currect values for pset already
+
+    Args:
+        vparams (dict): Parameter dict
+        grids (list): List of grids
+        surveys (list): List of surveys
```
cwjames1983, line 1639: I think passing the surveys might be historical - originally information was different between the two, maybe we remove passing surveys if no info is needed anymore?

cwjames1983 (approved):
Looks all good! Maybe I made a few use**** comments, but please go ahead and merge

zdm/craco/MC_Surveys/CRACO_alpha1_Planck18_Gamma.dat
profxj: fixed.  and for the future too (I hope).

zdm/grid.py
profxj: updated

zdm/iteration.py
profxj: Since the `survey` is held by the `grid`, it is superfluous.
But am keeping it around for now.

PR #13, opened 2022-02-02
Speedup igamma

profxj:
Code to speed up the incomplete gamma function using a
simple spline approach.

Tested in a Notebook and on the `submini` grid for `craco`


zdm/grid.py
```diff
@@ -69,15 +68,16 @@ def __init__(self, survey, state,
...
+            self.vector_cum_lf=energetics.vector_cum_power_law
+            self.array_diff_lf=energetics.array_diff_power_law
+            self.vector_diff_lf=energetics.vector_diff_power_law
         elif self.luminosity_function==1:  # Gamma function
-            self.array_cum_lf=zdm.array_cum_gamma
-            self.vector_cum_lf=zdm.vector_cum_gamma
-            self.array_diff_lf=zdm.array_diff_gamma
-            self.vector_diff_lf=zdm.vector_diff_gamma
+            self.array_cum_lf=energetics.array_cum_gamma_spline
+            self.vector_cum_lf=energetics.vector_cum_gamma
```
cwjames1983, line 77: should there not be an equivalent vector_cum_gamma_spline function here? After all, the 2D version is mostly a wrapper around the 1D version

cwjames1983 (changes requested):
Hi X, awesome you have done the spline speedup!

However, possible bug - see comment on grid.py. Will approve as soon as this is resolved.

cwjames1983 (approved):
All good! Even if "Explore_mini.ipynb" is a mystery to me :-)

PR #15, opened 2022-02-15
Running CRACO mini

profxj:
As titled


cwjames1983 (commented):
With the caveat that I never use ipython notebooks (for some reason I have had trouble getting them running on my laptop, should try again) and have no idea what a yaml files is - seems legit :-)

PR #16, opened 2022-02-20
Implemented scattering into FRB width calculation

cwjames1983:
So, I've switched around some parameter notation, and implemented a new method to include scattering in the FRB width. This was *definitely* a worthwhile thing to do (as in, referee should ask for it, and it will make no difference to our results).

Probably this will break something somewhere that directly uses parameters. Thankfully none of the affected parameters gets used in cubing so I hope we are safe there. Fingers crossed! ;-)


zdm/craco/loading.py
```diff
@@ -35,14 +35,22 @@ def set_state(alpha_method=1, cosmo=Planck18):
     vparams['FRBdemo']['source_evolution'] = 0
     
     vparams['beam'] = {}
-    vparams['beam']['thresh'] = 0
-    vparams['beam']['method'] = 2
+    vparams['beam']['Bthresh'] = 0
```
profxj, line 38: let's make sure this is updated everywhere! :)

zdm/cube.py
```diff
@@ -1,4 +1,4 @@
-
+#WARNING - THIS ROUTINE USES OUTDATED NOTATION AND NEEDS TO BE UPDATED
```
profxj, line 1: I'd say this file is fully DEPRECATED

zdm/parameters.py
```diff
@@ -22,19 +22,12 @@ class AnalysisParams(data_class.myDataClass):
 # Beam parameters
 @dataclass
 class BeamParams(data_class.myDataClass):
-    thresh: int = field(
-        default=0,
-        metadata={'help': '??'})
-    method: int = field(
+    Bmethod: int = field(
         default=2,
         metadata={'help': 'Method for calculation. See beams.py:simplify_beam() for options'})
```
profxj, line 27: This field is auto-magically slurped into the Table for the paper. 
  so would be nice to give explicity info when possible.

zdm/parameters.py
```diff
@@ -143,18 +136,58 @@ class HostParams(data_class.myDataClass):
...
-    logsigma: float = field(
+    Wlogsigma: float = field(
         default = 0.899148,
         metadata={'help': 'Intrinsic width log of sigma',
                   'unit': 'ms',
                   'Notation': '\sigma_{w}',
                   })
+    Wthresh: int = field(
+        default=0.5,
+        metadata={'help': 'Starting fraction of intrinsic width for histogramming'})
```
profxj, line 153: please add the `unit` for each of these.

zdm/survey.py
```diff
@@ -394,11 +397,10 @@ def find(self,lst, val):
             return [i for i, x in enumerate(lst) if x==val]
     
     def get_key(self,key):
-        print(self.keys)
-        print(self.keylist)
-        exit()
-        #if key in self.keys:
-        #    return self.keys
+        #print(self.keys)
```
profxj, line 400: Good catch!

zdm/survey.py
```diff
@@ -486,7 +488,53 @@ def calc_relative_sensitivity(DM_frb,DM,w,fbar,t_res,nu_res,model='Quadrature',d
...
-def make_widths(s:Survey,wlogmean,wlogsigma,nbins,scale=2,thresh=0.5):
+def geometric_lognormals(lmu1,ls1,lmu2,ls2,bins=None,Nrand=10000,plot=False,Nbins=101):
+    '''
+    Numerically evaluates the resulting distribution of y=\sqrt{x1^2+x2^2},
+    where logx1~normal and logx2~normal with log-mean lmu and 
+    log-sigma ls.
+    This is typically used for two log-normals of intrinsic
+    FRB width and scattering time
+    
+    lmu1, ls1: log mean and log-sigma of the first distribution
```
profxj, line 499: Can you specify the data type for each of the main inputs?
  Really helps.  Thx!

zdm/survey.py
```diff
@@ -592,15 +713,13 @@ def load_survey(survey_name:str, state:parameters.State, dmvals:np.ndarray,
     # Do it
     srvy=Survey()
     srvy.name = survey_name
-    srvy.process_survey_file(os.path.join(sdir, dfile), NFRB=NFRB, iFRB=iFRB)
+    path=os.path.join('/Users/cjames/CRAFT/FRB_library/Git/H0paper/zdm/data/Surveys/',dfile)
```
profxj, line 716: no absolute paths to `cjames`!  :)

zdm/tests/Scattering/test_scattering_methods.py
```diff
@@ -0,0 +1,125 @@
+
```
profxj, line 1: please move this file up one level to `tests`

I will then turn on CI to have the code auto-magically tested.

zdm/tests/Scattering/test_scattering_methods.py
```diff
@@ -0,0 +1,125 @@
...
+from zdm.craco import loading
+from zdm import io
+
+#from IPython import embed
+
+import numpy as np
+from zdm import survey
+from matplotlib import pyplot as plt
+
+def main():
```
profxj, line 27: name this something like `def test_scatt_blah`
  Needed for CI

profxj: And you may need to turn off the plotting..

profxj (changes requested):
Great stuff.

Made a few requests.  thx

zdm/craco/loading.py
cwjames1983: I wrote a script to grep for things like "'beam'" etc and it found some files under "papers" and "orig_tests". I guess you are using the "papers" section? But six further files got cleaned up and modded. I think this is everything. But if you like once this is merged back point me to your exact files and make sure

zdm/cube.py
cwjames1983: Yes, but wanted to make this explicit. CLearly to be removed at some point, e.g. in the cleanup once we merged back to main before the paper comes out.

zdm/parameters.py
cwjames1983: What is "the table"? I can certainly improve this (lots of the code needs better documentation) but this is tangential to this pull request, all I am doing here is renaming a thing

zdm/parameters.py
cwjames1983: Done. All without units are unitless, but have added empty field of ''

zdm/survey.py
cwjames1983: Oops, I know why this is there. It's because you implemented your darn if (name=thing) then (get it from ...) notation, and I needed to looks somewhere else! Forgot this temporary cheat was in there. I'll undo this. Can we undo this from the web page?

zdm/tests/Scattering/test_scattering_methods.py
cwjames1983: done! I think at least this is what you are after. Will be tough to reduce this to a simple yes/no output but will see.

zdm/tests/Scattering/test_scattering_methods.py
cwjames1983: Oh, oops. Plots are important. Anyway, added a tweak to toggle plotting on/off in program. Dodgy but oh well.

zdm/survey.py
cwjames1983: done.

zdm/tests/Scattering/test_scattering_methods.py
cwjames1983: Btw I did a 2D plot check of before/after, and everything is fine. The "new" method here is actually CHIME's parameters, which result in FRBs which are not as wide at even 892 MHz, hence the 'new' method results in more FRBs, particularly at low DM where DM smearing is not so important. However our measurements show wider FRBs than CHIME sees, hence I revert to our old width description PLUS scattering.

zdm/survey.py
cwjames1983: Unwound this now

cwjames1983:
OK, I have implemented all changes except related to "the table", where I think I need more input from you as to what you are after. Let me know

profxj (approved):
Great stuff.

And I'll deal with the Table (it is in the draft on Overleaf)

Thanks!

PR #17, opened 2022-02-21
Output updates + dolist=5

profxj:
Implements `dolist=5` for the cubes and also
  modifies the output to be simple CSV files.


cwjames1983 (approved):
Seems legit.

PR #20, opened 2022-02-27
Another mini CRACO run

profxj:
As titled

### DIFF
--- a/zdm/iteration.py
+++ b/zdm/iteration.py
@@ -5,6 +5,7 @@ import time
 from IPython.terminal.embed import embed
 import matplotlib.pyplot as plt
 import numpy as np
+import pandas
 from scipy.optimize import minimize
 # to hold one of these parameters constant, just remove it from the arg set here
 from zdm import cosmology as cos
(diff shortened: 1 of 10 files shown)


zdm/iteration.py
```diff
@@ -1032,15 +1028,14 @@ def cube_likelihoods(grids:list,surveys:list,
                 print("Iteration ",nth," took ",t2-t1," seconds")
 
             # write output: parameters, likelihoods (and components)
-            #f.write(string)
-            outputs.append(odict)
-            
-        #t1=time.process_time()
-        #print("Loop: ",i+run*howmany," took ", t1-t0," seconds")
-        #t0=t1
+            tdf = pandas.DataFrame([odict])
```
cwjames1983, line 1031: all good, but I'm not used to pandas, so hopefully it's all trivial to read in!

cwjames1983 (approved):
All good. I'm not familiar with pandas, so not 100% sure how easy/hard it will be to read objects in, but looks like it makes things simpler so no probs here.

profxj:
`pandas` is nothing special.  The output files are CSV.
  Pretty sure you know how to read those in...  ;)

PR #21, opened 2022-03-01
Prepping for Real cube

profxj:
As titled.  Let's check out the outputs before merging


cwjames1983 (commented):
I have snuck in some changes of my own here, e.g. see changes to iteration.py and the test_cube that asserts that this is correctly implemented.

Probably good to have you look at iteration.py and test_cube yourself just to check (good coding practice, but honestly maybe I should just approve this :-)

Otherwise everything is fine.

PR #24, opened 2022-04-04
Speedup

cwjames1983:
Very small set of changes that removes the likelihood calculation from the minimise constant only. I have tested this with the routine zdm/tests/Performance/test_update_performance.py and this works nicely. But since we want to use it asap (will save maybe... 30-40% of time?) I've made this a pull request rather than a simple push.


profxj (approved):
Looks good to me!

PR #25, opened 2022-04-15
Full real

cwjames1983:
I kept developing full real. Probably reasonable to leave main untouched...? In any case, I wanted to merge this in and then decide what I should keep developing on. Anyway, these are all minor changes.


profxj (approved):
merging

PR #27, opened 2022-04-24
New mini Real run

profxj:
A few files to generate the `mini Real` cube in the Cloud.


cwjames1983 (approved):
All good, to the extent that I know anything about the cloud submission commands.

PR #35, opened 2022-08-30
Linear + Log10 speedup

Eddysunday012:
# Description

To improve upon the speed up for `igrid.calc_pdv()`, we changed two main functionalities:

- Changing spline interpolation to linear interpolation
- Converting to log10 space

Both changes were measured separately and had different sets of functions that were modified. A full analysis can be found in `linear_profile.ipynb` under `zdm/tests/Performance` along with a summary under this google doc: 

https://docs.google.com/document/d/1iTCUdaC89qdxhUYQPOyFmZV0SrjRIHE6YBBfxktiUlc/edit?usp=sharing

# Fixes/Changes

Under `grid.py`
- in igrid.calc_pdv(), added log10 base conversion before traversing nested for loop
- in igrid.init, added use_log10 parameter (boolean) to control use of log10 conversion with its default value being false

Under `energetics.py`
- added dictionaries `igamma_linear` and `igamma_linear_log10` to account for linear and log10 interpolators
- added `vector_cum_gamma_linear()`, `array_cum_gamma_linear()`, and `init_igamma_linear()` for linear interpolation
- in `init_igamma_linear()` and `vector_cum_gamma_linear()`, added if else branch for log10 base conversion


zdm/energetics.py
```diff
@@ -19,6 +21,35 @@ def init_igamma_splines(gammas, reinit=False):
             # iGamma
             igamma_splines[gamma] = interpolate.splrep(avals, numer)
 
+def init_igamma_linear(gammas, reinit=False, log=False):
```
profxj, line 24: please add a doc string here.

zdm/energetics.py
```diff
@@ -19,6 +21,35 @@ def init_igamma_splines(gammas, reinit=False):
...
+
+                numer = np.array([float(mpmath.gammainc(
+                    gamma, a=iEE)) for iEE in avals])
+
+                # convert avals to log10 space (init x values)
+                log_avals = np.log10(avals)
+
+                igamma_linear_log10[gamma] = interpolate.interp1d(log_avals, numer)
+
+        else:
```
profxj, line 42: If the only difference in the two paths of the `if/else` is line `38` then can we reduce it to just that line?  Or maybe it needs to be that line and the one below it that fills the `dict`.

zdm/grid.py
```diff
@@ -41,6 +41,9 @@ def __init__(self, survey, state,
         self.b_fractions=None
         # State
         self.state = state
+        
+        # FOR TESTING WITH LOG 10
```
profxj, line 45: Is this comment still accurate?

zdm/grid.py
```diff
@@ -235,17 +240,32 @@ def calc_pdv(self,beam_b=None,beam_o=None):
             self.b_fractions=np.zeros([self.zvals.size,self.dmvals.size,self.beam_b.size])
         
         # for some arbitrary reason, we treat the beamshape slightly differently... no need to keep an intermediate product!
-        for i,b in enumerate(self.beam_b):
+        main_beam_b = self.beam_b
+        new_thresh = np.log10(self.thresholds) # use when calling in log10 space conversion
```
profxj, line 244: This might be expensive.  Can we only do it if `self.use_log10 is True` ?

zdm/grid.py
```diff
@@ -235,17 +240,32 @@ def calc_pdv(self,beam_b=None,beam_o=None):
...
+
+        # call log10 beam
+        if self.use_log10:
+            main_beam_b = np.log10(main_beam_b)
+
+        for i,b in enumerate(main_beam_b):
             for j,w in enumerate(self.eff_weights):
+                
+                # original
+                thresh = self.thresholds[j,:,:]/b
```
profxj, line 254: Shouldn't we skip this if `self.use_log10 is True`?

profxj (changes requested):
A few edits suggested.
Thanks @Eddysunday012 !

zdm/energetics.py
cwjames1983: Agree, looks like the if/else should be only around line 38

zdm/energetics.py
```diff
@@ -19,6 +21,35 @@ def init_igamma_splines(gammas, reinit=False):
             # iGamma
             igamma_splines[gamma] = interpolate.splrep(avals, numer)
 
+def init_igamma_linear(gammas, reinit=False, log=False):
+
```
cwjames1983, line 25: add quick doco line on log (yes, I know this was not originally documented, but we are slowly trying to remedy this!)

zdm/energetics.py
```diff
@@ -186,6 +217,47 @@ def vector_cum_gamma_spline(Eth:np.ndarray, *params):
     result[low]=1.
     return result
 
+def vector_cum_gamma_linear(Eth:np.ndarray, *params):
+    """ Calculate cumulative Gamma function using a linear interp1d
+
+    Args:
+        Eth (np.ndarray): [description]
```
cwjames1983, line 224: Eth: detection threshold energy in ergs
Return: cumulative probability above Eth

zdm/energetics.py
```diff
@@ -186,6 +217,47 @@ def vector_cum_gamma_spline(Eth:np.ndarray, *params):
...
+    """
+    params=np.array(params)
+    Emin=params[0]
+    Emax=params[1]
+    gamma=params[2]
+    log = params[3]
+
+    # Calculate
+    norm = float(mpmath.gammainc(gamma, a=Emin/Emax))
+    Emin_temp = Emin
```
cwjames1983, line 237: This gets redefined at line 245, i.e. this line is only used if linear. Shift this line to line 248

zdm/grid.py
cwjames1983: In particular, shouldn't the use_log10 be passed through the 'state'? Or is this now just a speedup which should work and hence we hard-code it?

zdm/tests/Performance/profile_linear.py
```diff
@@ -0,0 +1,126 @@
+""" Time profiling """
+
+######
+# first run this to generate surveys and parameter sets, by
```
cwjames1983, line 4: Add comment about the expected output: is this compute time in seconds for each method?

cwjames1983 (commented):
A few minor changes requested, typically a little bit of documentation, and moving some commands within if/else clauses

zdm/energetics.py
profxj: done

zdm/energetics.py
profxj: fixed

zdm/energetics.py
profxj: added

zdm/energetics.py
profxj: done

zdm/grid.py
profxj: fixed

zdm/grid.py
profxj: fixed

profxj (approved):
ready

PR #36, opened 2022-08-31
Paper resub

profxj:
This and that related to resubmitting.


cwjames1983 (approved):
Probably should test all this cleanly when it is committed... but what's there now looks fine

PR #37, opened 2022-10-11
Code and analysis related to 20220610A

profxj:
As title.

Mainly a script to run the Emax calculation.

I will add in the survey file in a moment too now that the source is public.


cwjames1983 (approved):
I think we should decide what to do with functions like "loading.py" where you have a temporary one-off hack. In the long term, they will get burdened with lot of bits and pieces of such hacks. Maybe create a special copy of the functions and put in a special paper directory?

PR #40, opened 2022-12-04
Refactor Survey class and files

profxj:
As titled.

Primary goal is to make the survey files machine-readable by
a "normal" database package.  Here, we make use of the ECSV
format of `astropy.table.Tables`

I stopped short of refactoring here there and everywhere.

Here are items to discuss/complete before merging:

To do/discuss:

-   Update all TNS
-   Confirm THRESH and SNRTHESH as `meta`
-   Confirm X columns are ok
-   Ok to ignore SNRth in Parkes?
-  Move forward with NewSurvey?


profxj:
Ok, @cwjames1983 --  Following our chat last night I did the following things:

- Made `SNRTHRESH` and `THRESH` properties of each FRB
- Modified `grid.py` accordingly (we currently take only one value)
- Merged in `main` which I should have done earlier.  Dealt with some conflicts
- Commented out the ~5 FRBs in the Parkes Survey file which had `MISSING` for `SNRth`
- Rename `SNRth` -> `SNRTHRESH` in the Parkes file.

This is now ready for your careful review.

zdm/survey.py
```diff
@@ -556,14 +806,10 @@ def geometric_lognormals(lmu1,ls1,lmu2,ls2,bins=None,
     
 def make_widths(s:Survey,state):
     """
+    WHAT DOES THIS METHOD DO?
```
cwjames1983, line 809: This method takes a distribution of intrinsic FRB widths (lognormal, defined by wlogmean and wlogsigma), and returns a list of w_i, p(w_i), where the w_i are i=1...N values of width, and p(w_i) are statistical weights associated with each. The \sum_i p(w_i) should sum to unity always. Each w_i is used to calculate a separate efficiency table.

zdm/survey.py
```diff
@@ -713,44 +970,160 @@ def load_survey(survey_name:str, state:parameters.State, dmvals:np.ndarray,
     """
     print(f"Loading survey: {survey_name}")
     if sdir is None:
-        sdir = os.path.join(resource_filename('zdm', 'data'), 'Surveys')
+        sdir = os.path.join(
+            resource_filename('zdm', 'data'), 'Surveys')
+        if original:
+            sdir = os.path.join(sdir, 'Original')
 
     # Hard code real surveys
```
cwjames1983, line 978: Hi @profxj I've been thinking about a few survey parameters, in particular nbins for beams, and how to parameterise efficiency. I *think* I've concluded that nbins needs to either go in the survey file itself, or at least we should allow for a default value there that can be over-written explicitly. In any case, I'd like to chat about this at some point, and need to think a bit more about what CHIME needs. This is of course fine for now, but it might need to be changed

zdm/survey.py
```diff
@@ -713,44 +970,160 @@ def load_survey(survey_name:str, state:parameters.State, dmvals:np.ndarray,
...
-    srvy=Survey()
-    srvy.name = survey_name
-    srvy.process_survey_file(os.path.join(sdir, dfile), NFRB=NFRB, iFRB=iFRB)
-    #srvy.process_survey_file(os.path.join(sdir, dfile), NFRB=NFRB, iFRB=iFRB)
-    srvy.init_DMEG(state.MW.DMhalo)
-    srvy.init_beam(nbins=Nbeams, method=state.beam.Bmethod, plot=False,
-                thresh=state.beam.Bthresh) # tells the survey to use the beam file
-    pwidths,pprobs=make_widths(srvy,state)
-    _ = srvy.get_efficiency_from_wlist(dmvals,pwidths,pprobs)
+    if original:
```
cwjames1983, line 1011: Thanks for keeping original functionality. At some point I expect we can drop backwards compatibility, but great to have it for now.

cwjames1983 (approved):
Hey X, this is a really nice refactor. And thanks for including backwards compatibility!

I don't know if this one question you had will change anything prior to merging - suspect not. I think we might need to perform some moderate changes to the survey to allow for CHIME (I'm currently working out how best to implement this in a branch based on the old survey format) but this can wait until Jan 2023.

zdm/survey.py
profxj: updated

zdm/survey.py
profxj: `nbins` now part of the data model.

We may want an analysis section for such parameters

PR #41, opened 2023-02-02
Branch to reproduce results from GW-FRB pulication in Nature Astronomy

cwjames1983:
This branch adds to "papers" to provide code for Nature Astronomy publication of GW190425 / FRB190425A.

Don't merge this in yet - I still need to add some files, but first need to check which ones I'm adding (confirming with GW people).

But we plan on putting in the final editor version paper tomorrow, so we need to add a code access statement. Hence I want this to be reviewed now.

But NOTE: none of the core code has changed, this is entirely an addition to the 'papers' directory.


profxj (approved):
Well documented and impressive set of code.

I wonder if some of it should be migrated into the 
main code base as it could have general value.  But
I'll let @cwjames1983 decide when/how.

profxj (approved):
Looks great!

PR #45, opened 2023-11-16
Efficiencies and MCMC branches

JordanHoffmann3:
Functionality used in my paper which allows 'eta' (SNR efficiency as a function of DM) to be read directly from a .npy file.

MCMC wrapper for the code.

Added uDMG and allowed DMhalo to vary.

Created some basic scripts for DSA files. Also added FAST survey.


zdm/MCMC.py
```diff
@@ -0,0 +1,141 @@
...
+import time
+
+import multiprocessing as mp
+
+from zdm.misc_functions import *
+
+#==============================================================================
+
+def mcmc_likelihoods(outfile:str, walkers:int, steps:int, params, surveys, grids, ncpus=1):
+    posterior_sample = mcmc_runner(calc_log_posterior, outfile, params, surveys, grids, nwalkers=walkers, nsteps=steps, ncpus=ncpus)
```
cwjames1983, line 27: Jordan: can you please add some doc string for these functions? E.g. format of:
"""
[documentation]
"""
Specifically, give an overview of the function, then describe what the inputs are for? E.g. is "walkers" the integer number of walkers? Is there a suggested number (O~1, 10, 100...)? etc.

See some other functions for descriptions. Not everything here is fully documented, but I think this file is a pretty important one to specify as fully as possible.

zdm/MCMC.py
```diff
@@ -0,0 +1,141 @@
...
+    print("Total time taken: " + str(end - start))
+    
+    posterior_sample = sampler.get_chain()
+
+    return posterior_sample
+
+#==============================================================================
+
+def get_likelihood(grid, s):
+
```
cwjames1983, line 118: Great function! This probably appears in several places, right? I think this function should appear in iteration.py, since it's got more general functionality than MCMC. But happy to keep it here for the moment

zdm/MCMC_wrap.py
```diff
@@ -0,0 +1,114 @@
...
+        for name in names:
+            filename = 'data/Surveys/' + name
+            s=survey.Survey(state, name, filename, dmvals, edir='data/Efficiencies/')
+
+            surveys.append(s)
+    
+        # Initialise grids
+        grids=initialise_grids(surveys,zDMgrid, zvals,dmvals,state,wdist=True)
+
+        # Save surveys / grids in pickle format
```
cwjames1983, line 71: In my experience, saving surveys via Pickle is not worth it - they take up a LOT of space, and loading them might only save a minute at most. I can imagine this accidentally producing 5GB of data on somebody's computer. Could you wrap this behaviour with a bool that says "if [pickle_it] then ...", and have this bool default to false?

zdm/craco/loading.py
```diff
@@ -95,7 +95,7 @@ def survey_and_grid(survey_name:str='CRAFT/CRACO_1_5000',
             state_dict=None, iFRB:int=0,
                alpha_method=1, NFRB:int=100, 
                lum_func:int=2,sdir=None,nz=500,ndm=1400,
-               nbins=5):
+               nbins=5,edir=''):
     """ Load up a survey and grid for a CRACO mock dataset
 
     Args:
```
cwjames1983, line 101: Can you please add a description of "edir" to the "Args" directory, and also add this functionality to "real_loading.py" in the main zdm directory?

zdm/parameters.py
```diff
@@ -146,18 +146,21 @@ class RepeatParams(data_class.myDataClass):
...
 class MWParams(data_class.myDataClass):
     ISM: float = field(
-        default=35.0,
-        metadata={"help": "Assumed DM for the Galactic ISM in units of pc/cm^3"},
-    )
+        default=35.,
+        metadata={'help': 'Assumed DM for the Galactic ISM',
+                  'unit': 'pc cm$^{-3}$',
+        })
+    uDMG: float = field(
```
cwjames1983, line 153: How annoying would it be to replace "u" with "sigma"? The reason is that "u" is awfully like "\mu", which is a mean. If this only appears O~10 times in the code, it's be great to rename this to sigmaDMG... (we also use "sigma" for host galaxy uncertainty)

zdm/scripts/plot_pzdm_grid.py
```diff
@@ -23,7 +23,7 @@
 def main():
```
cwjames1983, line 24: general comment: this script seems to have been altered for just specific purposes. In general, don't commit changes to such files which are simply minor alterations for testing purposes. In any case, what's here is fine, since I think I need to make a "simple" and "complex" version of this, and I'd left a bunch of annoying functionality in this to begin with, i.e. I hadn't followed my own advice!

zdm/scripts/run_mcmc.slurm
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+#SBATCH --job-name=CRAFT
+#SBATCH --output=../mcmc/Hoffmann2023_CRAFT_no_191228.out
+#SBATCH --ntasks=20
+#SBATCH --time=24:00:00
```
cwjames1983, line 5: Does it actually take 24 hr? Looks like this has just been set to the max possible. Or... is the run time genuinely close to 24 hr? (e.g. 20 hr or so?)

zdm/scripts/run_mcmc.slurm
```diff
@@ -0,0 +1,60 @@
...
+source $ZDM/.zdm_env/bin/activate
+
+cd $ZDM/zdm
+
+# outfile="Hoffmann2023_CRAFT_no_191228"
+outfile="DSA"
+walkers=40
+steps=1000
+
+surveys="DSA3.ecsv" #"parkes_mb_class_I_and_II.ecsv" # DSA3.ecsv FAST.ecsv CRAFT_class_I_and_II.ecsv private_CRAFT_ICS_892.ecsv private_CRAFT_ICS_1300.ecsv private_CRAFT_ICS_1632.ecsv
```
cwjames1983, line 22: Just to be sure: this current implementation makes it look like *only* DSA is being modelled by default. Can you make sure the version being committed has "everything" as the default behaviour, and other, commented alternatives have notes like "used for HOFFMAN2023" or similar? Or if there are old junk options here you'd never revisit, just delete them?

zdm/survey.py
```diff
@@ -495,7 +498,8 @@ def __init__(self, state, survey_name:str,
         self.process_survey_file(filename, NFRB, iFRB)
         # DM EG
         self.init_DMEG(state.MW.DMhalo)
-        
+        # Zs
+        self.init_zs()
```
cwjames1983, line 502: I feel like there should be some option to turn this on or off. The reason is that it's possible to have missing redshifts, but for good reasons - e.g., one simply has not yet got an optical follow-up. For ASKAP, I expect this option will kill lots of real FRBs. Let's discuss this, but I don't think we want this here, operating in the background all the time

zdm/survey.py
```diff
@@ -509,12 +513,59 @@ def __init__(self, state, survey_name:str,
...
     def init_DMEG(self,DMhalo):
         """ Calculates extragalactic DMs assuming halo DM """
         self.DMhalo=DMhalo
         self.DMEGs=self.DMs-self.DMGs-DMhalo
+        # self.DMEGs_obs=self.DMs-self.DMGs-DMhalo
+        # self.DMEGs = np.copy(self.DMEGs_obs)
+        # self.DMEGs[self.DMEGs < 0] = 10. # Minimum value of 10. pc/cm^3
+    
+    def init_zs(self):
+        """Gets zlist and nozlist and determines which z values to use"""
```
cwjames1983, line 527: See comments. Keep this functions, but we need to alter how it is used

cwjames1983: Btw, have you checked this works when there are no FRBs in the file? I've been tripped up by this before...

zdm/survey.py
```diff
@@ -971,7 +1026,7 @@ def make_widths(s:Survey,state):
         # this is done for tests, for complex surveys such as CHIME,
         # or for estimating the properties of a single FRB
         weights.append(1.)
-        widths.append(np.exp(slogmean))
```
cwjames1983, line 974: I'm noting the question about "should this be wlogmean". Actually, yes, it "should", although usually that won't matter at all. Can you please change it and remove this #NOTE ?

cwjames1983 (commented):
Super nice work Jordan! A few comments, and maybe one important discussion to have. We can chat more tomorrow on this issue (see SLACK message for now to get you thinking).

zdm/scripts/run_mcmc.slurm
JordanHoffmann3: I run it with enough walkers / steps that it caps out at 24 hours... generally it is better to run with more steps for longer because you don't lose as much to burn-in, then your parallelization comes with more walkers.

cwjames1983 (approved):
Nice work Jordan!

cwjames1983 (approved):
Debugging...

cwjames1983 (approved):
Approved again, let's see if this passes the tests!

PR #47, opened 2024-01-09
Fixed spline interp problem

cwjames1983:
Some minor changes to the behaviour of spline interpolation (interp in log space now), and changed 1e-2 to 1e-5 for a "small" number in survey.py. Also a spline performance test routine added. See SLACK for details.


zdm/energetics.py
```diff
@@ -10,20 +10,236 @@
...
+                smallest = numer[ismall]
+                numer[zero] = smallest
+                lnumer = np.log10(numer)
+                igamma_splines[gamma] = interpolate.splrep(lavals, lnumer,k=k)
+            else:
+                igamma_splines[gamma] = interpolate.splrep(avals, numer,k=k)
+            
+
+
+def test_spline_accuracy(gamma = -1.5837, Ntest = 100):
```
profxj, line 51: please move this to a file in the tests/ folder

profxj: extra credit if you get it to run under CI

zdm/energetics.py
```diff
@@ -10,20 +10,236 @@
...
+    
+    plt.sca(ax3)
+    plt.legend()
+    plt.ylim(1e-14,1)
+    plt.tight_layout()
+    plt.savefig('spline_rel_errors.pdf')
+    plt.close()
+    
 
+def time_splines(gamma = -1.5837, Ntimetest = 100000, Nreps=100):
```
profxj, line 165: probably best in tests/ too

profxj (changes requested):
One semi-significant change requested by me

Thanks for this!

zdm/energetics.py
cwjames1983: Done

zdm/energetics.py
cwjames1983: Done! In same script

cwjames1983: (in theory, I could split this into a performance check file, but too lazy, since we haven't really paid attention to performance checks in any systematic way - should re-do these at some point)

profxj (approved):
nice!

PR #48, opened 2024-01-11
Fixed log10 issue in repeat_grid.py

cwjames1983:
I forgot to modify this file to account for spline interpolation in log space!!!!! Thanks and apologies to Jordan for picking this up.


profxj:
whoops! :)

One quick thing -- why didn't the test catch it?

PR #54, opened 2024-07-18
Vvmax

cwjames1983:
This is for Wayne's V/Vmax paper: see papers/Vvmax

It uses FRB redshifts from the ICS paper.

I've pulled in main - there were, amazingly, no conflicts to resolve!

"zdm" is only used for calculating, well, z(DM), and also luminosity distances, but it was a very convenient library for doing so.


papers/Vvmax/Data/all_unloc_frb_data.dat
```diff
@@ -0,0 +1,64 @@
+# FRB                 DM        Freq          S/N    S/Nth   Dmg    dt         w         theta_d     Fnu       Zmac Nants RF
```
profxj, line 1: Is this data that others will want to use for future analyses?

if so, let's "expose" it by having it in `zdm/data/`

papers/Vvmax/RLFs.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
```
profxj, line 1: similar to the data, if any of the code that follows will be useful to the community, it should be in the main repo folder, e.g.

`zdm/vvmax/RLFs.py`

papers/Vvmax/RLFs.py
```diff
@@ -0,0 +1,125 @@
...
+# normalisation energy - no physical meaning, just makes units more sensible
+E0=1e22
+
+
+def make_first_guess(E,L, params,method):
+    """
+    Makes a first guess by setting the amplitude
+    to that of the first bin
+    
+    Params are always gamma,Emax
```
profxj, line 14: define `E, L, method`
  include data type if you can

papers/Vvmax/RLFs.py
```diff
@@ -0,0 +1,125 @@
...
+        mag = power_law(E[0],1.,params[0],params[1])
+        guess = [L[0]/mag,params[0],params[1]]
+    
+    
+    return guess
+    
+    
+
+
+def logIntegrateSchechter(log10E,A,gamma,log10Emax):
```
profxj, line 36: I'm not going to nit pick the doc strings unless these wind up 
in the main Repo folder (i.e. `zdm`).  

So if you move them there (it seems you probably should), then 
add short descriptions of each input paraemter

papers/Vvmax/calc_vvmax.py
```diff
@@ -0,0 +1,764 @@
+"""
+Script to calculate V and Vmax for CRAFT FRBs
```
profxj, line 2: is the code generic enough to be ueseful for other experiments?

if so, all the more reason to have this exposed in the main repo

papers/Vvmax/calc_vvmax.py
```diff
@@ -0,0 +1,764 @@
...
+
+# initialises cosmology and distance measures 
+
+
+def main(fname,outfile,Nsfr,alpha=0.,loc=False,minz=None,write=True,
+    laststrings=None,LimZ=None,zfrac = 0.):
+    """
+    
+    """
+    global macquartz
```
profxj, line 31: `global`s are so...  lazy  ;)

papers/Vvmax/calc_vvmax.py
```diff
@@ -0,0 +1,764 @@
...
+    # i.e. it ignores the fact the intrinsic width behaves differently...
+    # eff0 = (1+z0)**-0.5
+    # effz = (1+ztrials)**-0.5
+    
+    sz = s0 * (effz/eff0) * modFz
+    
+    return sz,eff0,effz,modFz
+    
+
+def calc_zmax(s0,z0,DM0,DMcosmic0,w0,freq,tsamp,state,Mz,MDM,alpha):
```
profxj, line 330: this looks like a very useful method for the community
   there is a somewhat similar one in the `FRB` repo already

papers/Vvmax/calc_vvmax.py
```diff
@@ -0,0 +1,764 @@
...
+    plt.scatter(DM, Fluences/Fnu)
+    plt.xlabel("DM")
+    plt.ylabel("James F / Arcus F")
+    plt.ylim(0,2)
+    #plt.xscale("log")
+    #plt.yscale("log")
+    plt.savefig("frb_fluences.pdf")
+    plt.close()
+
+def calc_fluence(F0,Bvals,SNR,w,tsamp,DM,freq,Nants):
```
profxj, line 488: holy cow, this module is large

maybe break it up if you move code from it into `zdm`

papers/Vvmax/calc_vvmax.py
```diff
@@ -0,0 +1,764 @@
...
+    BeamThetaDeg = data[:,8]
+    oldFnu = data[:,9]
+    zDM = data[:,10]
+    Nants = data[:,11]
+    newFnu = data[:,12]
+    return names,DM,freqMHz,SNR,SNRth,DMG,tsamp,width,BeamThetaDeg,newFnu,zDM,Nants
+
+
+
+global macquartz
```
profxj, line 684: a `global` here, a `global` there, makes code hard to follow and debug..

papers/Vvmax/make_paper_cdf_plot.py
```diff
@@ -0,0 +1,109 @@
+"""
```
profxj, line 1: I haven't reviewed the plots.

profxj (changes requested):
please consider how much of the code could be exposed in the main folder of the
repo, e.g .in `zdm/vvmax`

otherwise, looks great!

papers/Vvmax/Data/all_unloc_frb_data.dat
cwjames1983: OK, I'll have a think about that. Ideally, one might open a survey file and do V/Vmax for that survey. I don't think I'll implement that however. But certainly some routines would be useful. Good point, I'll check it out, and maybe re-factorise some routines.

cwjames1983: I've shifted some routines to zdm for vvmax - notably, the actual calculations of V and Vmax! - but the data is in a pretty unique format. I think I'll keeo it that way

papers/Vvmax/RLFs.py
cwjames1983: I don't think these ones have much value. Mostly, RLFs.py
- keeps constants from the literature, but only for three papers, and only a few constants
- has some specialised routines that integrate the functions over histogram bins
Obviously, they could be useful, but they're also of rather marginal use.

papers/Vvmax/Data/all_unloc_frb_data.dat
cwjames1983: So see new file vvmax.py in the zdm directory

papers/Vvmax/RLFs.py
cwjames1983: I've added proper doc strings to the routines that have been shifted to zdm/vvmax.py . Or, at least, *better* doc strings!

papers/Vvmax/calc_vvmax.py
cwjames1983: Yep, this is the part that's been moved. At least, the core of it has.

papers/Vvmax/calc_vvmax.py
cwjames1983: Yeah I know. Will tell you privately some of this history of this at some point

papers/Vvmax/calc_vvmax.py
cwjames1983: Yep, it's in the main code now, sort-of kind0of properly documented

papers/Vvmax/make_paper_cdf_plot.py
cwjames1983: I guess they're in the paper :-)

papers/Vvmax/RLFs.py
cwjames1983: Done

cwjames1983: I have defined many of these common values at the top level, since they are used throughout the file

papers/Vvmax/calc_vvmax.py
cwjames1983: I think I've appropriately broken up useful functions from this in the main module, under vvmax.py.

papers/Vvmax/calc_vvmax.py
cwjames1983: No globals in the part of the code in the main repo!

profxj (approved):
excellente

PR #59, opened 2024-10-23
171020

cwjames1983:
This is legacy code from maybe 2019. It won't run, it's poorly commented, etc. But I don't have the time to go back and recreate it for the purpose of posterity. Nonetheless, I think it's useful to preserve it. So - here it is!


profxj (approved):
totally fine;  thx for adding it in

PR #63, opened 2025-01-30
Rep debug

cwjames1983:
Main fix: debugged an error in how the expected number of repeaters and single bursts were calculated. See /zdm/scripts/scan_rgamma_likelihood.py for a demo of this now working.

Main new feature: a routine inside repeaters.py to calculate the probability of a repeating FRB having Nrepeats given is is a repeater. see scripts /zdm/scripts/calculate_repeater_probability.py


JordanHoffmann3 (approved):
Looks good, glad to finally have that bug fixed!

PR #64, opened 2025-04-29
Mcmc reps

JordanHoffmann3:
Merging in changes for DMhalo, MCMC updates and repetition updates.


papers/MCMC/Surveys/DSA.ecsv
```diff
@@ -0,0 +1,50 @@
+# %ECSV 1.0
```
profxj, line 1: should this be in `zdm/data/Surveys` ?

[same for FAST below]

papers/MeerTRAP/incoherent.py
```diff
@@ -0,0 +1,267 @@
...
+
+def main():
+    
+    # in case you wish to switch to another output directory
+    
+    opdir='Incoherent/'
+    
+    # approximate best-fit values from recent analysis
+    # best-fit from Jordan et al
+    if True:
```
profxj, line 47: this should propagate to the inner code base somehow.
  
Maybe, the call `state = parameters.State()`
  should now accept an optional reference, e.g. Hoffman+2025

papers/MeerTRAP/plot_z_comparison.py
```diff
@@ -0,0 +1,248 @@
+"""
```
profxj, line 1: I love this figure

zdm/dmg_sanskriti2020.py
```diff
@@ -0,0 +1,103 @@
+import numpy as np
```
profxj, line 1: add a note saying what this module is for

zdm/iteration.py
```diff
@@ -105,29 +105,6 @@ def get_names(which=None):
         names=names[which]
     return names
 
-def print_pset(pset):
```
profxj, line 108: am not sad to see this go :)

zdm/parameters.py
```diff
@@ -19,6 +19,20 @@ class AnalysisParams(data_class.myDataClass):
...
+        default=30,
+        metadata={
+            "help": "Discard FRBs below this absolute galactic latitude",
+            "unit": "Degrees"
+        }
+    )
+    DMG_cut: float = field(
+        default=None,
+        metadata={
+            "help": "Discard FRBs above this DMG",
```
profxj, line 32: and is `DMG` the Galactic DM estiamted from NE2001?

profxj (approved):
wow, that's an incredible amount of excellent work

just a few bits and pieces from me to fuss with

papers/MCMC/Surveys/DSA.ecsv
cwjames1983: @JordanHoffmann3 ? I tend to agree with X. If this is for a previous DSA sample - e.g., one with only a few FRBs - then maybe we create a new directory under "Surveys" titled "Historical"? But then again, if this is now a deprecated version of DSA that contained only some of their FRBs I think it's OK to stay here

papers/MeerTRAP/incoherent.py
cwjames1983: @profxj  Yes. This is something that I think Jordan and I should implement: whenever we publish something, we record both best-fit state parameters as an optional argument, and record a large sample of walkers to be randomly sampled from.

papers/MeerTRAP/plot_z_comparison.py
cwjames1983: It is beautiful, isn't it? :-) I really like the colour scheme!

papers/Repeaters/utilities.py
```diff
@@ -194,16 +194,16 @@ def get_chime_data(DMhalo=50,snrcut=None):
     chimedir = 'CHIME_FRBs/'
     infile = chimedir+'chimefrbcat1.csv'
     
-    idec=5
```
cwjames1983, line 197: What changed here? I am a bit puzzled by the increase in the number of FRBs to 600. The data sample for this paper definitely only included 536.

zdm/data/DM_combined.txt
```diff
@@ -0,0 +1,78 @@
+#l	b	DM	e_l	e_u	DMmax	e_sys
+303.48	-6.42	60.71335927	6.323289758	6.323289758	67.03664903	0
```
cwjames1983, line 2: This doesn't look like a file that should be in the repository. What is "DM_combined"? Is this something for a specific analysis? I assume so, in which case, this should be moved to a directory for a particular paper

zdm/data/Surveys/MC_CRAFT_ICS_1300.ecsv
cwjames1983, line 1: Is such a file something we need to keep in the repository? After all, an MC is by definition able to be regenerated. In general, data like this I'd suggest not to add to git. However, if it's important to have a fixed sample because this is what was used for a specific work, then I'm perfectly happy to keep it, but maybe move it to the particular "paper" directory for that?

zdm/data/Surveys/MC_CRAFT_ICS_1300_2.ecsv
cwjames1983, line 1: Ditto with this

zdm/scripts/run_mc_sample.slurm
cwjames1983, line 1: I think that slurm scripts are sufficiently specialised that they should not live here. One might put this in a zdm/scripts/slurm directory perhaps, which contains slurm commands for launching jobs?

zdm/scripts/run_mcmc.slurm
cwjames1983, line 1: Ditto as above

papers/FitRepetition2025/ICS_Survey_Analysis/ICS_complex.txt
cwjames1983, line 1: Just wanted to note that this is a record of all ICS observations. It's actually quite relevant to the ICS survey paper Ryan produced

zdm/survey.py
```diff
@@ -528,92 +539,109 @@ def init_repeaters(self):
...
-                elif self.TOBS is None:
-                    self.TOBS = self.Nfields * self.Tfield
-                else:
-                    # All three are set
-                    print("TOBS, TFIELD and NFIELDS all specified")
-                    self.Tfield = self.TOBS / self.Nfields
-                
-                print("set TOBS = " + str(self.TOBS) + ", TFIELD = " + str(self.Tfield) + ", NFIELDS = " + str(self.Nfields))
+        
+        # checks to see if NREP info is present. If not, assumes all single bursts
```
cwjames1983, line 543: This might be deprecated now, since NREP is a required column with a default value of 1. Will see later though

zdm/survey.py
```diff
@@ -816,24 +896,64 @@ def process_survey_file(self,filename:str,
...
+                        frbs.append(frb)
+                    else:
+                        excluded += 1
+                elif np.abs(frb['Gb']) > min_lat:
+                    frbs.append(frb)
+                else:
+                    excluded += 1
+            
+            print("Using minimum galactic latitude of " + str(min_lat) + ". Excluding " + str(excluded) + " FRBs")
+        # Max DM
```
cwjames1983, line 951: Should this also increment the "excluded" count? I see a DMG cut acting just like the Gb cut

zdm/survey.py
```diff
@@ -852,6 +973,9 @@ def process_survey_file(self,filename:str,
             self.meta['WIDTH'] = np.median(self.frbs['WIDTH'])
             self.meta['DMG'] = np.mean(self.frbs['DMG'])
         
+        # fills in missing coordinates is possible
```
cwjames1983, line 976: I should move this to before the cuts section, since fixing coordinates to get correct cut values for Gb (i.e. in RA and DEC are set)

cwjames1983 (changes requested):
Huge series of changes here. A few to-do items, but nothing major. Note however to pull in early in future.

papers/Repeaters/utilities.py
JordanHoffmann3: I'm not sure, I do not have any recollection of changing that...

zdm/data/DM_combined.txt
JordanHoffmann3: These are the observations for the Sanskriti+ DMG method, so they are needed for that galactic model

zdm/data/Surveys/MC_CRAFT_ICS_1300.ecsv
JordanHoffmann3: This was used for some analysis for the MCMC paper, but nothing we ended up putting into the paper. We can probably remove it?

zdm/scripts/run_mc_sample.slurm
JordanHoffmann3: Yep, moved to a slurm directory

zdm/survey.py
JordanHoffmann3: This defaults them to 1 anyway, but you're right it will be redundant in that sense

zdm/survey.py
JordanHoffmann3: Yes so if it is excluded based up DMG or Gb then we increment the excluded count

papers/Repeaters/utilities.py
cwjames1983: OK, let's leave this as is, we can always revert

zdm/data/DM_combined.txt
cwjames1983: Ah, so this is an input file for that calculation? Perhaps move this to a subdirectory of /data, even if it's called "/data/Misc", just so we don't get random files floating in data without much context. Perhaps also change the name to "Sanskriti_DM_inputs" or something a bit more descriptive?

zdm/data/Surveys/MC_CRAFT_ICS_1300.ecsv
cwjames1983: Yes, please remove it. If it was used for the paper, it should be in the papers directory.

zdm/survey.py
cwjames1983: Let's leave in for now - we might need to do a bit of a code clean at some point anyway

zdm/survey.py
cwjames1983: OK, please add the excluded count to the DMG test as well as the Gb test

zdm/survey.py
cwjames1983: OK, I have done this now

zdm/data/DM_combined.txt
JordanHoffmann3: Done

cwjames1983 (approved):
All good, thanks Jordan!

PR #70, opened 2025-05-08
Cleanup

cwjames1983:
This "cleanup" branch has performed the following changes:
- removed deprecated functionality associated with old_survey class
- removed unused routines from misc_functions and iteration
- shifted plot_grid_2 from misc_functions to plotting, renamed plot_grid, and changed these function calls in the various scripts
- re-organised the scripts folder into sub-folders according to script functionality.

No new functionality should be changed, except a couple more niceties to plot_grid such as only plotting the legend if it actually has legend entries to plot.


JordanHoffmann3 (approved):
Seems to be all sorted now!

profxj (approved):
Excellent;  thanks for all of this!

PR #74, opened 2025-11-03
A bunch of things! Most notably, explicit use of survey beamwidth

cwjames1983:
This branch has been a bit of a catch-all for a few miscellaneous updates being implemented simultaneously.

The most significant is the addition of explicit use of beam value B in calc likelihoods 1D and 2D, and the ability of survey.py to read in beam values internally. It's hoped that using this from even a single survey will allow explicit values of FRB energy to be used in constraining the luminosity function.

At the same time, I have revamped the use of "dolist" within calc_likelihoods_[1/2]D to report components of the likelihood. This is great for debug purposes - see script zdm/scripts/likelihood_components.py for examples of its use.

There are also maybe a few additions to scripts within the MeerTRAP and SKA "papers" folders, adding minor functionality.

And I did a few other bits of cleanup - deleted two legacy routines from iteration.py, got rid of some more resource_filename calls, and maybe fiddled here and there with other things.

### DIFF
--- a/zdm/survey_data.py
+++ b/zdm/survey_data.py
@@ -97,7 +97,7 @@ class FRB(data_class.myDataClass):
                   'Notation': '',
                   })
     WIDTH: float = field( 
-        default=0.1, 
+        default=-1, 
         metadata={'help': "Width of the event", 
                   'unit': 'ms', 
                   'Notation': '',


zdm/states.py
```diff
@@ -0,0 +1,378 @@
+"""
```
profxj: this is a great addition!

PR #81, opened 2026-04-08
Repeaters

JordanHoffmann3:
Mostly bugfixing to include repeaters. Seems to be working now, but gives very flat gamma in the luminosity function when including all available data.


papers/MeerTRAP/plot_z_comparison.py
cwjames1983, line 1: Hey @JordanHoffmann3 I see you have a lot of old code commented out in this file. Perhaps you can just delete it? You are in the best position to know if it's something that might be useful in the future - but if so, then maybe add a short note on what the commented code does? But I suspect it's a straight delete operation.

zdm/iteration.py
cwjames1983, line 1: I suspect that merging this with the current main will be an utter pain. I might download this branch myself and do the merging, then upload, since this could get pretty messy I think. Well, maybe.

cwjames1983 (commented):
Hey Jordan - great set of updates! My only real request is to get rid of old/commented-out text under plot_z_comparison, which seems to have a lot of junk in it.

Otherwise - great! let me know when you've done this, and I'll merge. And then pull it into my current branch, and resolve any/all conflicts.

Nice work!

cwjames1983 (approved):
I have just checked through the MeerTRAP part - not too bad, I'll approve this!

